PR #7, opened 2023-02-22
OPSEXP-2003 Bump alfresco-common with updatecli

gionn:
Ref: OPSEXP-2003

### DIFF
--- a/.github/workflows/updatecli.yaml
+++ b/.github/workflows/updatecli.yaml
@@ -35,7 +35,7 @@ jobs:
         run: updatecli diff
 
       - name: Run Updatecli in apply mode
-        if: github.event_name == 'push'
+        if: github.event_name == 'workflow_dispatch'
         env:
           UPDATECLI_GITHUB_TOKEN: ${{ secrets.BOT_GITHUB_TOKEN }}
           GIT_AUTHOR_EMAIL: ${{ vars.BOT_GITHUB_EMAIL }}
--- a/.github/workflows/updatecli.yaml
+++ b/.github/workflows/updatecli.yaml
@@ -1,0 +1,43 @@
+---
+name: Bump versions
+
+on:
+  workflow_dispatch:
+  push:
+    paths:
+      - '.github/workflows/updatecli.yaml'
+      - 'updatecli.d/**'
+    branches:
+      - 'OPSEXP-**'
+
+permissions:
+  contents: "write"
+  pull-requests: "write"
+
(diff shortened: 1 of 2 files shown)


alxgomz (approved):

alxgomz (commented):
We need to decide on a way to trigger these updates and which policy to apply. In particular why would we want to do a patch increment rather than a minor or a major

gionn:
> We need to decide on a way to trigger these updates and which policy to apply. In particular why would we want to do a patch increment rather than a minor or a major

consider this is mostly a draft to start having some automation, for the time being I was thinking to just have workflow_dispatch and run it on-demand: in the end we know when we release a new dependency.

if we start to add 3rd party dependencies support then we may want to add schedule.

another approach could be to hook to the workflow that updates github pages (so hopefully we are sure at that point the registry is updated)

.github/workflows/updatecli.yaml
```diff
@@ -0,0 +1,43 @@
...
+name: Bump versions
+
+on:
+  workflow_dispatch:
+  push:
+    paths:
+      - '.github/workflows/updatecli.yaml'
+      - 'updatecli.d/**'
+    branches:
+      - 'OPSEXP-**'
```
alxgomz, line 11: I'm not sure I understand the trigger logic here.
Should we not trigger this =for `paths: [chats/alfresco-common/**]` as well? I thin kthere will be cases where we do not (or forget) update the updatecli config/templates.
If in doubt maybe we should make it a workflow dispatch only with mandatory inputs `targets.*.spec.versionincrement`

gionn: `push` is just for testing changes to updatecli configuration, see that it runs only in diff mode
 https://github.com/Alfresco/alfresco-helm-charts/pull/7/files#diff-f3c1fa6e77c6c2264c2ed7ac4c908d4baee1aca7f4d2582914ec08138be8088cR30

while `workflow_dispatch` is the way I intended to use it (run in apply mode)
https://github.com/Alfresco/alfresco-helm-charts/pull/7/files#diff-f3c1fa6e77c6c2264c2ed7ac4c908d4baee1aca7f4d2582914ec08138be8088cR38

I thought to add the input for versionincrement but for now I think is enough, given that:
> I suppose that the automation is for covering trivial updates while if we do major rewrites/feature we'll bump manually within a PR.

alxgomz (approved):

PR #17, opened 2023-03-06
OPSEXP-1853: move named template back in chart

alxgomz:
Ref: OPSEXP-1853

### DIFF
--- a/charts/alfresco-search-service/templates/_helpers.tpl
+++ b/charts/alfresco-search-service/templates/_helpers.tpl
@@ -17,6 +17,17 @@ If release name contains chart name it will be used as a full name.
 {{- end }}
 
 {{/*
+Alfresco Search Service Host
+*/}}
+{{- define "alfresco-search.host" -}}
+{{- if index .Values.enabled -}}
+  {{ printf "%s-solr" (include "alfresco-search.fullName" .) -}}
+{{- else -}}
+  {{ index .Values "external" "host" | default "localhost" -}}
+{{- end -}}
+{{- end -}}
+
+{{/*
 Get Alfresco Search container Port ("internal")
 */}}
 {{- define "alfresco-search.containerPort" -}}
--- a/charts/alfresco-search-service/templates/_helpers.tpl
+++ b/charts/alfresco-search-service/templates/_helpers.tpl
@@ -20,11 +20,7 @@ If release name contains chart name it will be used as a full name.
 Alfresco Search Service Host
 */}}
 {{- define "alfresco-search.host" -}}
-{{- if index .Values.enabled -}}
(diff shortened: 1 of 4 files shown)


charts/alfresco-search-service/values.yaml
```diff
@@ -12,7 +12,6 @@ nodeSelector: {}
 # alfrescoRegistryPullSecrets has to be set
 # More information can be found on
 # https://github.com/Alfresco/acs-deployment/blob/master/docs/helm/registry-authentication.md
-type: search-services
```
gionn, line 15: what are we trying to achieve here?

L5 document this field

charts/alfresco-search-service/templates/_helpers.tpl
```diff
@@ -31,7 +42,11 @@ Get Alfresco Search container Port ("internal")
 Get Alfresco Search Service Port
 */}}
 {{- define "alfresco-search.svcPort" -}}
-{{- .Values.service.externallPort | default 80 | int -}}
+{{- $svcPort := 80 }}
+{{- if hasKey .Values "service" }}
+  {{- $svcPort := default 80 .Values.service.externalPort }}
```
gionn, line 47: no need to default, it's already 80

charts/alfresco-search-service/templates/_helpers.tpl
```diff
@@ -16,6 +16,17 @@ If release name contains chart name it will be used as a full name.
 {{- end }}
 {{- end }}
 
+{{/*
+Alfresco Search Service Host
+*/}}
+{{- define "alfresco-search.host" -}}
+{{- if index .Values.enabled -}}
```
gionn, line 23: what is this top level enabled that I don't see in the values file?

gionn (changes requested):

charts/alfresco-search-service/templates/_helpers.tpl
alxgomz: true that's a blind copy/paste from alfresco-common

charts/alfresco-search-service/templates/_helpers.tpl
alxgomz: well I still want to cover for cases where `.Values.service.externalPort` is not set, but I guess you'll prefer this other way of doing things

charts/alfresco-search-service/values.yaml
alxgomz: From what I could see this is not used anywhere

alxgomz: I thought it was unused and missed it actually still was

gionn (approved):
But please have a proper PR title

PR #20, opened 2023-03-13
OPSEXP-1854: introduce independent sync service chart

alxgomz:
Ref: OPSEXP-1854

Also adds kustomize to setup ci env for some charts where required

### DIFF
--- a/.github/workflows/lint-test.yaml
+++ b/.github/workflows/lint-test.yaml
@@ -73,7 +73,7 @@ jobs:
           kubectl create secret generic quay-registry-secret --from-file=.dockerconfigjson="${HOME}"/.docker/config.json --type=kubernetes.io/dockerconfigjson -n "$INSTALL_NAMESPACE"
           kubectl create secret generic broker-secret --from-literal=BROKER_URL="failover:(nio://activemq:61616)?timeout=3000" --from-literal=BROKER_USERNAME=admin --from-literal=BROKER_PASSWORD=admin -n "$INSTALL_NAMESPACE"
 
-      - name: prepare charts for testing
+      - name: Prepare charts mocks for testing
         if: steps.list-changed.outputs.changed == 'true'
         run: |
           for C in $(ct lsc --config ct-lint.yaml); do
--- a/charts/alfresco-sync-service/values.yaml
+++ b/charts/alfresco-sync-service/values.yaml
@@ -59,6 +59,7 @@ messageBroker:
     # failover:(nio://acs-activemq-broker:61616)?timeout=3000&jms.useCompression=true
     user: alfresco
     password: admin
+    # -- An existing kubernetes secret with MQ info (prefered over using values)
     existingSecretName:
   # Activemq chart options (see [Alfresco ActiveMQ chart
   # documentation](https://github.com/Alfresco/alfresco-helm-charts/tree/main/charts/activemq))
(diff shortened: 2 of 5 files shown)


charts/alfresco-sync-service/ci/kustomize/kustomization.yaml
```diff
@@ -0,0 +1,18 @@
...
+    files:
+      - httpd.conf
+
+helmChartInflationGenerator:
+  - chartName: apache
+    chartRepoUrl: https://charts.bitnami.com/bitnami
+    chartVersion: 9.2.18
+    releaseName: dummy-repo
+    releaseNamespace: alfresco
+    values: values.yaml
```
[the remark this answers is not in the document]
alxgomz: Things to consider:
* we already have [pre-requisites](https://github.com/Alfresco/alfresco-helm-charts/blob/632758d9c285321b367976393cc2fd825c65a5a4/.github/workflows/lint-test.yaml#L74) in the workflow.
* Here we're essentially trying to workaround a k8s unfriedly behavior of sync service (the real fix would be to have sync-service not fails miserably if it can't find a repo at startup).
The other options I considered:
* add that as a dependency of the chart => having that dummy stuff in a chart is really 
* adding a true repo chart dependency => we don't have such chart right now & implementing it is way beyond the scope of the ticket

charts/alfresco-sync-service/values.yaml
```diff
@@ -55,11 +55,28 @@ repository:
...
-  user: admin
-  password: admin
-  existingSecretName: null
+  enabled: false
+  external:
+    url:
+    # failover:(nio://acs-activemq-broker:61616)?timeout=3000&jms.useCompression=true
+    user: alfresco
+    password: admin
+    existingSecretName:
```
gionn, line 62: I am not sure about having extraneous values in a section of a dependency, that could have unexpected behavior in case of conflicts. And for what? just to have a single section in the values for handling everything related to an infra component (internal or external) - may not be bad for the user but now I am concerned about the consistency of this approach (we really need to document this soft design choice somewhere or we will endup doing inconsistent things with time passing)

.github/workflows/lint-test.yaml
```diff
@@ -73,6 +73,16 @@ jobs:
           kubectl create secret generic quay-registry-secret --from-file=.dockerconfigjson="${HOME}"/.docker/config.json --type=kubernetes.io/dockerconfigjson -n "$INSTALL_NAMESPACE"
           kubectl create secret generic broker-secret --from-literal=BROKER_URL="failover:(nio://activemq:61616)?timeout=3000" --from-literal=BROKER_USERNAME=admin --from-literal=BROKER_PASSWORD=admin -n "$INSTALL_NAMESPACE"
 
+      - name: prepare charts for testing
```
gionn, line 76:
```suggestion
      - name: Prepare charts mocks for testing
```

charts/alfresco-sync-service/values.yaml
alxgomz: That is very true.
I quickly created https://alfresco.atlassian.net/browse/OPSEXP-2043 to track that

alxgomz: If you don't mind, I will kepp this one for documentation purpose as it's not linked to another chart and remove the ones which are documented as part of the dependent charts

PR #22, opened 2023-03-14
OPSEXP-1954: need an acs-deployment friendly fallback url

alxgomz:
Ref: OPSEXP-1854

In acs-deployment we need to keep the messageBroker disabled (as syncservice needs to connect to the repo events)
but we can't pass a connection string as it's dynamically evaluated by the activemq chart in acs-deployment
Also adds the `.global` fallback mechanism


charts/alfresco-sync-service/templates/_helpers-mq.tpl
```diff
@@ -2,13 +2,13 @@
...
 {{- $brokerPort := .Values.messageBroker.services.broker.ports.external.openwire | int }}
+{{- if .Values.messageBroker.enabled }}
 {{- $brokerHostname := include "activemq.fullname" (dict "Values" .Values.messageBroker "Chart" .Chart "Release" .Release) }}
-{{- coalesce .Values.messageBroker.url (printf "failover:(nio://%s-broker:%d)%s" $brokerHostname $brokerPort $brokerOptions) }}
+{{- printf "failover:(nio://%s-broker:%d)%s" $brokerHostname $brokerPort $brokerOptions }}
 {{- else }}
-{{- required "Please provide a failover URL when using external message broker" .Values.messageBroker.external.url -}}
+{{- coalesce .Values.messageBroker.external.url (printf "failover:(nio://%s-%s-broker:%d)%s" .Release.Name .Values.messageBroker.nameOverride $brokerPort $brokerOptions) -}}
 {{- end }}
 {{- end -}}
```
gionn, line 13: behavior change without test changes means we need more tests

gionn (approved):

PR #23, opened 2023-03-16
OPSPEXP-1854: add a repo helper usable from parent chart

alxgomz:
acs-deployment approach requires to be able to expand repo service name from release name

### DIFF
--- a/charts/alfresco-sync-service/templates/_helpers.tpl
+++ b/charts/alfresco-sync-service/templates/_helpers.tpl
@@ -18,3 +18,11 @@ chart: {{ .Chart.Name }}
 {{ include "syncservice.selectorLabels" . }}
 heritage: {{ .Release.Service }}
 {{- end }}
+
+{{- define "syncservice.repository" -}}
+{{- if .Values.repository.nameOverride }}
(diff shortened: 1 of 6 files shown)


charts/alfresco-sync-service/templates/_helpers.tpl
```diff
@@ -18,3 +18,11 @@ chart: {{ .Chart.Name }}
...
 heritage: {{ .Release.Service }}
 {{- end }}
+
+{{- define "syncservice.repository" -}}
+{{- if .Values.repository.nameOverride }}
+{{- printf "%s-%s" .Release.Name .Values.repository.nameOverride | trunc 63 | trimSuffix "-" }}
+{{- else }}
+{{- .Values.repository.host }}
+{{- end }}
+{{- end -}}
```
gionn, line 28: we will move this once repository get its own chart I imagine

gionn (approved):

charts/alfresco-sync-service/templates/_helpers.tpl
alxgomz: I would like to yes but it mostly depends on what the acs-deployment charts becomes eventually

PR #24, opened 2023-03-16
OPSEXP-1854: Add missing service label selector

alxgomz:
Ref: OPSEXP-1854

### DIFF
--- a/charts/alfresco-sync-service/Chart.yaml
+++ b/charts/alfresco-sync-service/Chart.yaml
@@ -8,7 +8,7 @@ keywords:
 name: alfresco-sync-service
 sources:
   - https://github.com/Alfresco/acs-deployment
-version: 4.0.2
+version: 4.0.3
 appVersion: 4.0.0-M7
 icon: https://avatars0.githubusercontent.com/u/391127?s=200&v=4
 dependencies:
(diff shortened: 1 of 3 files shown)


charts/alfresco-sync-service/values.yaml
```diff
@@ -105,7 +105,10 @@ postgresql:
   primary:
     extendedConfiguration: |
       shared_buffers = 256MB
-      max_connections = 80
+      max_connections = 100
+      wal_level = minimal
+      max_wal_senders = 0
+      max_replication_slots = 0
```
gionn, line 111: what are those changes unrelated to this PR? 👀

gionn (changes requested):

charts/alfresco-sync-service/values.yaml
alxgomz: these are meant to provide more sensible default config for a "convenience" backend PG and avoid loosing precious connexion slot by reserving them for replication feature that will anyway not be used. This also allow to have a backend that has a configuration that do not requires 16GB RAM like before (with default settings and 400 connections slots allowed)

gionn: ok but why in a PR that fix a completely different problem and that will hide this change once it get merged to master?

alxgomz: Mostly to avoid releasing 2 versions for these 2 different issues (my tests showed the acs-deployment pipeline may be impacted).
Though I'm happy to to split it into 2 PR. But I see little value

gionn (approved):

PR #30, opened 2023-04-07
Made the ingress annotations variable & fixed secret in the JAVA_OPTION_TOOLS

shazada:


shazada:
And fixed the secret which was missing in the JAVA_OPTION_TOOLS hence the secret wasn't asked in the login.

charts/alfresco-search-service/templates/ingress.yaml
```diff
@@ -24,6 +23,9 @@ spec:
     {{- end }}
   {{- end }}
   rules:
+  {{- if .Values.ingress.hostName }}
+  - host: {{ tpl .Values.ingress.hostName . }}
+  {{- else }}
```
alxgomz, line 28: This `else` doesn't look to have a matching closing `end`, but I would even argue that one may want to specify both the hostname to bind the ingress on together with a `HTTPIngressRuleValue`

```suggestion
  {{- end }}
```

charts/alfresco-search-service/templates/secret-solr-jtoolopt.yaml
```diff
@@ -8,5 +8,5 @@ metadata:
     release: {{ .Release.Name }}
 type: Opaque
 data:
-  JAVA_TOOL_OPTIONS: {{ printf "-Dalfresco.secureComms.secret=%s " (include "tracking-shared-secret" .) | b64enc | quote }}
+  JAVA_TOOL_OPTIONS: {{ printf "-Dalfresco.secureComms=secret -Dalfresco.secureComms.secret=%s " (include "tracking-shared-secret" .) | b64enc | quote }}
```
alxgomz, line 11: Not sure why this is needed. This is actually dealt with in the configMap and the image's entrypoint:
* https://github.com/Contezza/alfresco-helm-charts/blob/5489feb0d069b30cce67ec45fb38298f00aa4b71/charts/alfresco-search-service/templates/config.yaml#L21
* https://github.com/Alfresco/SearchServices/blob/fe640439ada0bd531be7a2f53e5b8693b396392b/search-services/packaging/src/docker/search_config_setup.sh#L95

alxgomz (changes requested):
the secret in JAVA_TOOL_OPTION is unrelated to the basic auth setup by the ingress

charts/alfresco-search-service/templates/secret-solr-jtoolopt.yaml
shazada: Without it it doesn't work, i can login in SOLR without putting a secret

shazada: And it's included in the Docker-compose:
https://github.com/Alfresco/acs-deployment/blob/master/docker-compose/7.3.N-docker-compose.yml#L140

charts/alfresco-search-service/templates/ingress.yaml
shazada: I'll fix the else to end like suggested

alxgomz: Not a big deal, that's just the name of the object used in the rule: https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.19/#httpingressrulevalue-v1-networking-k8s-io

alxgomz:
I've pushed What I think are the correct changes in another branch (plus some slight improvements): https://github.com/Alfresco/alfresco-helm-charts/tree/Contezza/main

Feel free to cherry pick what suits you and test it. If that doesn't work for you would you please share reproduction steps?

charts/alfresco-search-service/templates/secret-solr-jtoolopt.yaml
shazada: I'm not sure, but when I deploy the search helm-chart with let's say only the following value file:
`global:
  tracking:
    auth: secret
    sharedsecret: secret
alfresco-search-service:
  repository:
    # The value for "host" is the name of this chart
    host: alfresco-alfresco-cs-repository
  searchServicesImage:
    # tag: 1.4.3.4
    tag: 2.0.6`

My config map only shows ALFRESCO_SECURE_COMMS: "secret"
And JAVA_TOOLS_OPTIONS is missing -Dalfresco.secureComms.secret=secret 

And i'm not even enabling the ingress, just the service and doing a portforward on the services to see if SOLR is running fine.

alxgomz: I have tested the following values (I could only assume the indentation) against the [branch](https://github.com/Alfresco/alfresco-helm-charts/tree/Contezza/main) I mentioned earlier:

```yaml
---
global:
  tracking:
    auth: secret
    sharedsecret: secret
alfresco-search-service:
repository:
  host: alfresco-alfresco-cs-repository
searchServicesImage:
  # tag: 1.4.3.4
  tag: 2.0.6
```

The resulting manifest has:
* a secret containing the JAVA_TOOL_OPTIONS containing the base64 encoded string `-Dalfresco.secureComms.secret=secret`
* a configmap which has the following variable: `ALFRESCO_SECURE_COMMS: secret`

As you can see o nthe screenshot below, this is properly reflected to the containers env and solr config file
image

> Note: this is a bot missleading as the authentication mechanism and the shared secret are the same: `secret`

Also trying to access the ingresss (after adding the following helm options `--set ingress.enabled=true --set ingress.hostName=localhost --set ingress.basicAuth='YWRtaW46JGFwcjEkVVJqb29uS00kSEMuS1EwVkRScFpwSHB2a3JwTDd1Lg=='` requires to have both curl options to enable basic auth & the shared secret header:

```sh
curl -H "X-Alfresco-Search-Secret: secret" -u admin:admin http://localhost/solr/
```

shazada: WTF so maybe the issue is in our own solr 2.0.6 image OMG let me revert to default. 
Yeah the ident are right when you push the search-services chart and then include it in another chart.
I was trying it like that, but will directly try you'r branch

shazada: Ok I tried it with the following values file:
`global:
  tracking:
    auth: secret
    sharedsecret: secret

ingress:
  enabled: true
  hostName: localhost
  basicAuth: YWRtaW46JGFwcjEkVVJqb29uS00kSEMuS1EwVkRScFpwSHB2a3JwTDd1Lg==

searchServicesImage:
  repository: alfresco/alfresco-search-services
  # tag: 1.4.3.4
  tag: 2.0.6
`

PS C:\Projecten\alfresco\alfresco-helm-charts\charts\alfresco-search-service> curl -H "X-Alfresco-Search-Secret: secret" -u admin:admin http://localhost/solr/
curl: (7) Failed to connect to localhost port 80 after 2235 ms: Couldn't connect to server

If I do a port-forward, then it works.
image

alxgomz:
Superseded by #36

PR #31, opened 2023-04-12
test renovate

gionn:


gionn:
lack support for automatic bump of chart version, better stick to updatecli

PR #35, opened 2023-04-13
OPSEXP-2072 Rewrite sync service mock with a simple testing subchart

gionn:
Ref: OPSEXP-2072


gionn:
Looks like pulling dependencies from a subchart is not the way to go

Trying with a full fledged chart in https://github.com/Alfresco/alfresco-helm-charts/pull/37

PR #37, opened 2023-04-17
OPSEXP-2072 Create alfresco-mocks chart

gionn:
Ref: OPSEXP-2072


alxgomz:
The main problem i see with this is that all packaged charts which depends on this mock chart (even conditionally) will ship the mock chart and its dependencies.
I prefer we have a hack in the CI jobs rather than in the deliverables.

gionn:
I am fine in not having an additional chart to release that is just for testing but I will keep this branch alive in case we want to come up later on this, I am still afraid that the current approach is not straightforward to debug and doesn't follow the standard development workflow

PR #38, opened 2023-04-18
OPSEXP-1862 Add alfresco repository chart

gionn:
Ref: OPSEXP-1862

- [ ] https://github.com/Alfresco/acs-deployment/issues/930


charts/alfresco-repository/templates/_helpers-name.tpl
gionn, line 1: please review first this

charts/alfresco-repository/templates/_helpers-label.tpl
gionn, line 1: please review first this approach

alxgomz (commented):
The naming system will indeed trigger a lot of changes in the other charts to be aligned.
However I think by being conservative we would be sticking too much to what we have today and we no is not optimal..
So that's ok for me to proceed with this new approach.
We'll probably need a dedicated Eepic for aligning charts...

gionn:
I will wait for https://github.com/Alfresco/acs-deployment/pull/902 and integrating the changes before continuing but worth another round of review

alxgomz (commented):
Mostly commented on the naming logic

charts/alfresco-repository/values.yaml
```diff
@@ -0,0 +1,397 @@
+repository:
```
alxgomz, line 1: That level of values should be removed and its content moved to the root context.

alxgomz (commented):
Beyond the comment, I am concerned the chart still contains a lot of logic that's inhereted from `alfresco-content-services` chart and shoudl only be present in the parent chart.
For example, all the logic related to configuring a service based on wether it's in-cluster or not (`activemq` vs `messageBroker`, `database` vs `postgresql`)

gionn:
Deprecated by https://github.com/Alfresco/alfresco-helm-charts/pull/84

PR #57, opened 2023-06-26
OPSEXP-2173 Fixup autocommit action with updatecli workflow

gionn:
Ref: OPSEXP-2173

### DIFF
--- a/.github/workflows/updatecli.yaml
+++ b/.github/workflows/updatecli.yaml
@@ -37,6 +37,6 @@ jobs:
         uses: stefanzweifel/git-auto-commit-action@v4.16.0
         with:
           commit_message: |
-            🛠 Updatecli pipeline bump #${{ github.event.pull_request.number }}
+            🛠 Updatecli pipeline bump
           branch: ${{ github.event_name == 'workflow_dispatch' && 'updatecli-bump' || '' }}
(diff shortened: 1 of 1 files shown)


.github/workflows/updatecli.yaml
```diff
@@ -27,12 +29,14 @@ jobs:
...
       - run: updatecli apply --experimental
         env:
           GITHUB_TOKEN: ${{secrets.GITHUB_TOKEN}}
       - name: Git Auto Commit
         uses: stefanzweifel/git-auto-commit-action@v4.16.0
         with:
           commit_message: |
-            🛠 Updatecli pipeline bump #${{ github.event.pull_request.number }}
+            🛠 Updatecli pipeline bump
+          branch: ${{ github.event_name == 'workflow_dispatch' && 'updatecli-bump' || '' }}
```
alxgomz, line 41: I think using an empty string for an input value does not trigger the default in the called action. Can you check that?

alxgomz (changes requested):

.github/workflows/updatecli.yaml
gionn: looks good, I tested previously adding synchronize event here - commits are not here anymore because I've throw them away while rebasing

alxgomz (approved):

PR #60, opened 2023-06-27
OPSEXP-2173 Add updatecli pipeline to bump acs components

gionn:
Ref: OPSEXP-2173

updatecli configs moved at https://github.com/Alfresco/alfresco-updatecli

### DIFF
--- a/.github/workflows/updatecli.yaml
+++ b/.github/workflows/updatecli.yaml
@@ -40,6 +40,7 @@ jobs:
             🛠 Updatecli pipeline bump
           branch: ${{ github.event_name == 'workflow_dispatch' && 'updatecli-bump-helm' || '' }}
           create_branch: ${{ github.event_name == 'workflow_dispatch' }}
+          push_options: ${{ github.event_name == 'workflow_dispatch' && '--force' || '' }}
 
   bump-acs-dependencies:
     runs-on: ubuntu-latest
--- a/.github/workflows/updatecli.yaml
+++ b/.github/workflows/updatecli.yaml
@@ -79,7 +79,7 @@ jobs:
         env:
           QUAY_USERNAME: ${{ secrets.QUAY_USERNAME }}
           QUAY_PASSWORD: ${{ secrets.QUAY_PASSWORD }}
-          UPDATECLI_GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          UPDATECLI_GITHUB_TOKEN: ${{ secrets.BOT_GITHUB_TOKEN }}
 
       - name: Git Auto Commit
         uses: stefanzweifel/git-auto-commit-action@v4.16.0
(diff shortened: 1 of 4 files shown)


.github/workflows/updatecli.yaml
```diff
@@ -38,6 +38,55 @@ jobs:
...
+      - name: Install Updatecli
+        uses: updatecli/updatecli-action@v2
+        with:
+          version: v0.54.0
+
+      - name: Checkout updatecli configs
+        uses: actions/checkout@v3
+        with:
+          repository: alfresco/alfresco-updatecli
+          ref: OPSEXP-2173-andromeda
```
gionn, line 71: revert to master after https://github.com/Alfresco/alfresco-updatecli/pull/1

alxgomz: Actually I like th idea to have "sprints" tags in this repo.

alxgomz (commented):
I generally like the approach but I'm only concerned with one thing:
the uber-manifest in the updatecli repo is very acs-deployment centric.
Wouldn't it be possible/better that this repo contains only scms, sources, condition, actions, ... (everything but targets) and each individual repo has the targets that match its own structure?

.github/workflows/updatecli.yaml
gionn: we can make a tag on alfresco-updatecli once everything is merged and seems working

gionn, line 71:
```suggestion
          ref: master
```

gionn:
> the uber-manifest in the updatecli repo is very acs-deployment centric.
> Wouldn't it be possible/better that this repo contains only scms, sources, condition, actions, ... (everything but targets) and each individual repo has the targets that match its own structure?

my current idea is that we keep a single updatecli manifest which needs to be flexible enough to handle the very different use case we have - but at least we a single place where to maintain/support version-related automations.

splitting the updatecli manifest is not really an option because it is not supported, we can only have a single templated manifest which behave differently depending on the values we provide (so each repo will have its own values file which will make the updatecli pipeline works as we want in that specific repo).

alxgomz:
> splitting the updatecli manifest is not really an option because it is not supported by the tools itself

Sure but what if we have a manifest that expects (or requires) the targets to be provided through a named template (`template` & `define`) ? That named template should be provided within the local repo and defines the expected targets while the sources still are maintained in the remote repo which remains a single source of truth....

alxgomz:
Disclaimer: just thinking out loud, I haven't any experimentation so far.

gionn:
> Sure but what if we have a manifest that expects (or requires) the targets to be provided through a named template (`template` & `define`) ?

I don't think it's an issue until we want to target something other than plain yml files but it's not I would think to solve now (supporting both acs-deployment and alfresco-helm-charts with a single config is already a good achievement for me)

alxgomz (approved):

PR #70, opened 2023-07-06
OPSEXP-2188: add nginx specific annotation to use from repo a share

alxgomz:
Ref: OPSEXP-2188

refactor to use from both share and repo

### DIFF
--- a/charts/alfresco-common/templates/_helpers-nginx.tpl
+++ b/charts/alfresco-common/templates/_helpers-nginx.tpl
@@ -12,6 +12,7 @@ Define annotations as provided in values
 
 {{/*
 Define required annotations for secure nginx ingress
+We're overridong here ANY server snippet annotation and applying only ours trusted ones
 */}}
 {{- define "alfresco-common.nginx.secure.annotations" }}
 {{- template "alfresco-common.nginx.annotations" . }}
--- a/charts/alfresco-common/templates/_helpers-nginx.tpl
+++ b/charts/alfresco-common/templates/_helpers-nginx.tpl
@@ -1,6 +1,6 @@
 {{/*
 Define annotations as provided in values
-Skip ANY server snippet annotation
+Skip ANY server-snippet annotation (CVE-2021-25742)
 */}}
 {{- define "alfresco-common.nginx.annotations" }}
 {{- range $annotation, $value := .ingress.annotations }}
@@ -12,8 +12,7 @@ Skip ANY server snippet annotation
 {{- end }}
 
 {{/*
-Define required annotations for secure nginx ingress
-Apply our trusted server snippet for Nginx
+Define required annotations for secure ACS/SHARE API access
 */}}
 {{- define "alfresco-common.nginx.secure.annotations" }}
     nginx.ingress.kubernetes.io/server-snippet: |
--- a/charts/alfresco-common/templates/_helpers-nginx.tpl
+++ b/charts/alfresco-common/templates/_helpers-nginx.tpl
@@ -1,5 +1,6 @@
 {{/*
 Define annotations as provided in values
+Skip ANY server snippet annotation
 */}}
 {{- define "alfresco-common.nginx.annotations" }}
 {{- range $annotation, $value := .ingress.annotations }}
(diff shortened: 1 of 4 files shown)


charts/alfresco-common/templates/_helpers-nginx.tpl
```diff
@@ -0,0 +1,24 @@
...
+  {{- if ne $annotation "nginx.ingress.kubernetes.io/server-snippet" }}
+  {{- $annotation | nindent 4 }}: |-
+    {{- $value | nindent 6 }}
+  {{- end }}
+{{- end }}
+{{- end }}
+
+{{/*
+Define required annotations for secure nginx ingress
+We're overridong here ANY server snippet annotation and applying only ours trusted ones
```
gionn, line 15:
```suggestion
Define required annotations for secure nginx ingress
We're overriding here ANY server snippet annotation and applying only ours trusted ones
```

charts/alfresco-common/templates/_helpers-nginx.tpl
```diff
@@ -0,0 +1,24 @@
+{{/*
+Define annotations as provided in values
+*/}}
+{{- define "alfresco-common.nginx.annotations" }}
+{{- range $annotation, $value := .ingress.annotations }}
+  {{- if ne $annotation "nginx.ingress.kubernetes.io/server-snippet" }}
```
gionn, line 6: we are also forbidding to use server-snippet here, we should probably document it here rather than in the other

charts/alfresco-common/templates/_helpers-nginx.tpl
```diff
@@ -0,0 +1,24 @@
...
+  {{- end }}
+{{- end }}
+{{- end }}
+
+{{/*
+Define required annotations for secure nginx ingress
+We're overridong here ANY server snippet annotation and applying only ours trusted ones
+*/}}
+{{- define "alfresco-common.nginx.secure.annotations" }}
+{{- template "alfresco-common.nginx.annotations" . }}
```
gionn, line 18: not totally sure it's a good idea to include the generic template here. What if we decide later that we need another template which add their own annotations? Better to keep also this one independent from the generic annotations template?

gionn (changes requested):

charts/alfresco-common/templates/_helpers-nginx.tpl
```diff
@@ -0,0 +1,24 @@
+{{/*
+Define annotations as provided in values
+Skip ANY server snippet annotation
```
gionn, line 3:
```suggestion
Skip ANY server-snippet annotation
```

why? security reasons? document it

charts/alfresco-common/templates/_helpers-nginx.tpl
```diff
@@ -0,0 +1,24 @@
...
+  {{- if ne $annotation "nginx.ingress.kubernetes.io/server-snippet" }}
+  {{- $annotation | nindent 4 }}: |-
+    {{- $value | nindent 6 }}
+  {{- end }}
+{{- end }}
+{{- end }}
+
+{{/*
+Define required annotations for secure nginx ingress
+Apply our trusted server snippet for Nginx
```
gionn, line 16: needs an update after the refactor

charts/alfresco-common/templates/_helpers-url.tpl
```diff
@@ -0,0 +1,45 @@
...
+{{- if kindIs "string" . }}
+  {{- $known_urls = splitList "," . }}
+{{- end }}
+{{- range $known_urls }}
+  {{- $parsed_url := urlParse . }}
+  {{- $known_url := urlJoin (dict "host" $parsed_url.host "path" $parsed_url.path "scheme" $parsed_url.scheme) }}
+  {{- $csrf_origins = append $csrf_origins $known_url }}
+{{- end }}
+{{- $csrf_origins | join "," }}
+{{- end -}}
```
gionn, line 33: isn't most of the code in common with `alfresco-common.csrf.referer`?

gionn (changes requested):

charts/alfresco-common/templates/_helpers-url.tpl
alxgomz: actually my definition of origin was wrong. The only part that's common is the url split which is not very interesting to turn into a named template of its own as it involves turning output into json and reading from json where it needs to be used (given template only output strings)

gionn: there is a chance to have some unit tests here? those functions are hard to understand what they do otherwise

alxgomz: not using `helm unittest` as it's a library chart. The consumer charts should implement these tests maybe?

alxgomz: Some unit tests in the share chart: https://github.com/Alfresco/alfresco-helm-charts/blob/6b4bdc011c41ad6069e3b64b4673d0ad036b4583/charts/alfresco-share/tests/cm_test.yaml#L15

gionn (approved):

PR #71, opened 2023-07-07
OPSEXP-2116 repository service named template alfresco-connector-ms365

slohe1:


gionn (approved):

charts/alfresco-connector-ms365/templates/_helpers.tpl
```diff
@@ -49,3 +49,11 @@ Selector labels
 app.kubernetes.io/name: {{ include "alfresco-connector-ms365.name" . }}
 app.kubernetes.io/instance: {{ .Release.Name }}
 {{- end }}
+
+{{- define "alfresco-connector-ms365.repository" -}}
```
gionn, line 53: you are not using it anywhere

add a test

gionn (changes requested):

alxgomz (changes requested):
the `nameOverride` or `fullnameOberride` for the repo should be set by the parent chart.
I have tried that same approach of using override in subchart and figured out it has several drawback. It creates a lot of nesting, and doesn't play well with external components (where you would have the subchart's component disabled but still leverage the override underneeth it...

I think a better approach is to expect the parent chart to pass the repo URL details in a string (e.g. using the `fullnameOverride`  passed in `alfresco-connector-ms365.repository.host`) and also support reading the repo details from an existing configmap (generated by the parent chart.

charts/alfresco-connector-ms365/values.yaml
```diff
@@ -43,3 +43,5 @@ repository:
   host: acs-alfresco-cs-repository
   # -- ACS repository port
   port: 80
+  # -- a pre-existing configmap which provides expected configuration for ms365
+  existingConfigMap:
```
gionn, line 47:
```suggestion
  existingConfigMap: null
```

or updatecli will complain later :D

gionn (approved):

PR #72, opened 2023-07-10
OPSEXP-2188: create a dedicated share chart

alxgomz:
Ref: OPSEXP-2188


charts/alfresco-share/Chart.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: v2
+name: alfresco-share
+description: Alfresco Share Helm chart for Kubernetes
+type: application
+version: 0.1.0-alpha.0
+appVersion: 7.4.0
+dependencies:
+  - repository: https://alfresco.github.io/alfresco-helm-charts
+    version: 2.1.0-alpha.0
```
gionn, line 9: remember after test to release also common (avoid having a stable chart using an unstable dep)

charts/alfresco-share/templates/NOTES.txt
gionn, line 1: we kept this file in other charts? I remember not

charts/alfresco-share/templates/deployment-share.yaml
```diff
@@ -0,0 +1,86 @@
+# Defines the deployment for the alfresco share app
```
gionn, line 1:
```suggestion
```

charts/alfresco-share/templates/service-share.yaml
```diff
@@ -0,0 +1,15 @@
+# Defines the service for the alfresco share app
```
gionn, line 1:
```suggestion
```

charts/alfresco-share/templates/serviceaccount.yaml
gionn, line 1: do we really need a service account?

gionn (changes requested):

charts/alfresco-share/templates/serviceaccount.yaml
alxgomz: it's a good practice and doesn't hurt

charts/alfresco-share/templates/NOTES.txt
alxgomz: yes but we made it more relevant. Thank you pointing it!

charts/alfresco-share/Chart.yaml
alxgomz: that chart is also alpha. Will do it only only stable release

gionn: I would not tie that to the sprint, just have the version to be a reflection of the current state we think the chart is  (after testing on our integration pipeline we release a stable after an alpha)

gionn (approved):

PR #78, opened 2023-07-19
OPSEXP-2188: missing ingress class in share

alxgomz:
Ref: OPSEXP-2188

### DIFF
--- a/charts/alfresco-share/templates/ingress.yaml
+++ b/charts/alfresco-share/templates/ingress.yaml
@@ -6,8 +6,7 @@ We only support nginx ingress for now: https://alfresco.atlassian.net/browse/OPS
 */}}
 {{- if not (semverCompare ">=1.18-0" .Capabilities.KubeVersion.GitVersion) }}
   {{- $_ := unset .Values.ingress.annotations "kubernetes.io/ingress.class" }}
-  {{- $_ := set .Values.ingress.annotations "kubernetes.io/ingress.class" "nginx" }}
-*/}}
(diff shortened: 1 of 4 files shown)


charts/alfresco-share/templates/ingress.yaml
```diff
@@ -6,8 +6,7 @@ We only support nginx ingress for now: https://alfresco.atlassian.net/browse/OPS
 */}}
 {{- if not (semverCompare ">=1.18-0" .Capabilities.KubeVersion.GitVersion) }}
   {{- $_ := unset .Values.ingress.annotations "kubernetes.io/ingress.class" }}
-  {{- $_ := set .Values.ingress.annotations "kubernetes.io/ingress.class" "nginx" }}
-*/}}
+  {{- $_ = set .Values.ingress.annotations "kubernetes.io/ingress.class" "nginx" }}
```
gionn, line 9: `:=`?

gionn (changes requested):

charts/alfresco-share/templates/ingress.yaml
alxgomz: Once the var has been defined and assigned using `:=` you can update it with a regular `=` assignment.
(Though both seem to work)

gionn (approved):

charts/alfresco-share/templates/ingress.yaml
alxgomz: source: https://go.dev/doc/go1.11#minor_library_changes (text/template)

PR #81, opened 2023-07-26
OPSEXP-2116 fixup alfresco-connector-ms365 chart

slohe1:


charts/alfresco-connector-ms365/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v2
 name: alfresco-connector-ms365
 description: A Helm chart for deploying Alfresco connector ms365 service
 type: application
-version: 0.3.0
+version: 0.4.0
```
alxgomz, line 5: That is a bugfix, please increment patch version, instead

alxgomz (changes requested):

alxgomz (approved):

PR #84, opened 2023-07-28
OPSEXP-1862: preliminary repository chart

alxgomz:
Ref: OPSEXP-1862

Initial implementation:
* deployment with support for:
  * persistence
  * db config
  * env vars
* configmap properties
* basic helm features


gionn (commented):
would you mind squashing all commits from the second to the last? <3 (so I get here a diff of only the changes between the initial import commit and yours)

alxgomz:
> would you mind squashing all commits from the second to the last? <3 (so I get here a diff of only the changes between the initial import commit and yours)

Done, though the "initial commit" may already contain some slight compared to a vanilla chart

charts/alfresco-repository/ci/kustomize/kustomization.yaml
```diff
@@ -0,0 +1,17 @@
+---
```
gionn, line 1: I am not a fan of this approach because is already causing issues in the sync chart because nothing is going to wipe the resources installed by this chart when the test finish and that's a problem when testing more than one chart and due to runners limited resources

charts/alfresco-repository/templates/configmap.yaml
```diff
@@ -0,0 +1,29 @@
...
+    -Ddb.password=${DATABASE_PASSWORD}
+    -Ddb.driver={{ include "alfresco-repository.db.driver" .Values.db }}
+    -Dencryption.keystore.type=JCEKS
+    -Dencryption.cipherAlgorithm=DESede/CBC/PKCS5Padding
+    -Dencryption.keyAlgorithm=DESede
+    -Dencryption.keystore.location=/usr/local/tomcat/shared/classes/alfresco/extension/keystore/keystore
+    -Dmetadata-keystore.aliases=metadata
+    -Dmetadata-keystore.metadata.algorithm=DESede
+    -Dmetadata-keystore.password=mp6yc0UD9e
+    -Dmetadata-keystore.metadata.password=oKIWzVdEdA
```
gionn, line 23: can we add a comment to state why we are hardcoding this (historical reasons?)

charts/alfresco-repository/tests/deployment_test.yaml
```diff
@@ -0,0 +1,150 @@
...
+#              secretName: somesecret
+#              defaultMode: 0400
+#        template: deployment.yaml
+#      - contains:
+#          path: spec.template.spec.containers[0].volumeMounts
+#          content:
+#            mountPath: >-
+#              /usr/local/tomcat/shared/classes/alfresco/extension/license/
+#            name: acs-license
+#        template: deployment.yaml
```
gionn, line 51: ?

charts/alfresco-repository/ci/kustomize/kustomization.yaml
```diff
@@ -0,0 +1,17 @@
...
+secretGenerator:
+  - name: alfsecrets
+    literals:
+      - DATABASE_USERNAME=alfresco
+      - DATABASE_PASSWORD=alfresco
+generatorOptions:
+  disableNameSuffixHash: true
+
+helmChartInflationGenerator:
+  - chartName: postgresql
```
gionn, line 12: in all the other charts we have migrated we have postgresql as optional dependency of the chart, stick to that

charts/alfresco-repository/ci/kustomize/kustomization.yaml
```diff
@@ -0,0 +1,17 @@
+---
+namespace: alfresco
+secretGenerator:
+  - name: alfsecrets
+    literals:
+      - DATABASE_USERNAME=alfresco
+      - DATABASE_PASSWORD=alfresco
```
gionn, line 7: this secret can be part of the chart, why force the user to provide a secret?

gionn (changes requested):

charts/alfresco-repository/ci/kustomize/kustomization.yaml
alxgomz: At that stage there is no values to pass db credentials. Hence the required secret

charts/alfresco-repository/ci/kustomize/kustomization.yaml
alxgomz: To be honest I like the optional dependecies even less for the reason explained below.
Though  I admitedly didn't take the reosource issue into account

charts/alfresco-repository/templates/configmap.yaml
alxgomz: Keystore configuration is something I want to address in a later PR so this one do not inflates even more

charts/alfresco-repository/ci/kustomize/kustomization.yaml
alxgomz: Tracked in https://alfresco.atlassian.net/browse/OPSEXP-2216 for now

charts/alfresco-repository/ci/kustomize/kustomization.yaml
alxgomz: Tracked i nhttps://alfresco.atlassian.net/browse/OPSEXP-2216 for now

charts/alfresco-repository/templates/NOTES.txt
gionn, line 1: remove this?

gionn (approved):

PR #88, opened 2023-08-01
OPSEXP-1862: add support for license

alxgomz:
Ref: OPSEXP-1862

### DIFF
--- a/charts/alfresco-repository/README.md
+++ b/charts/alfresco-repository/README.md
@@ -1,6 +1,6 @@
 # alfresco-repository
 
-![Version: 0.1.0-alpha.1](https://img.shields.io/badge/Version-0.1.0--alpha.1-informational?style=flat-square) ![Type: application](https://img.shields.io/badge/Type-application-informational?style=flat-square) ![AppVersion: 23.1.0-A21](https://img.shields.io/badge/AppVersion-23.1.0--A21-informational?style=flat-square)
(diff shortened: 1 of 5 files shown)


charts/alfresco-repository/README.md
```diff
@@ -19,7 +19,8 @@ Alfresco content repository Helm chart
 | args | list | `[]` |  |
 | command | list | `[]` |  |
 | configuration.repository.existingConfigMap | string | `nil` | a configmap containing the "alfresco-global.properties" key populated with actual Alfresco repository properties |
-| configuration.repository.existingSecret | string | `nil` | Name of a pre-existing secret TODO: secret documentation |
+| configuration.repository.existingSecret.keys.license | string | `"license.lic"` | Key within the secret that contains the license file (MUST end with ".lic") |
+| configuration.repository.existingSecret.name | string | `"repository-secrets"` | Name of a multi-purpose pre-existing secret |
```
gionn, line 23: I like the pattern where we can specify both secret name and keys inside a secret, but in this way we are like enforcing to always use a single secret.

it should be more flexible if we have something like:

```
configuration.repository.license.existingSecretName
configuration.repository.license.existingSecretKey
```

and so on for every feature that requires a secret

so it's up to the user if using a single fat secret referencing the same secret name in every existingSecretName or multiple secrets or a combination of both

gionn (changes requested):

charts/alfresco-repository/README.md
alxgomz: I actually did it on purpose to not let user creating a mess of plumbing secrets/config to vars/volumes together.
With the number of different variables availables in acs I'm afraid it because unmanageable while we can "more easily" maintain a list of secret/configmap available for people to use in specific are of the product (e.g.: one for mail config, one for db, one for....)

gionn: this last one is probably too more generic, I am more afraid to the fact that even for use that we manage license with an external script and the rest with terraform, it would be a problem to manage a single secret if we have multiple keys for multiples secret into it

gionn (approved):

PR #89, opened 2023-08-01
OPSEXP-1862: add message broker support

alxgomz:
Ref: OPSEXP-1862

### DIFF
--- a/charts/alfresco-repository/templates/configmap.yaml
+++ b/charts/alfresco-repository/templates/configmap.yaml
@@ -21,6 +21,7 @@ data:
     -Dmetadata-keystore.metadata.algorithm=DESede
     -Dmetadata-keystore.password=mp6yc0UD9e
     -Dmetadata-keystore.metadata.password=oKIWzVdEdA
+    -Dmessaging.broker.url=${BROKER_URL}
     -Dmessaging.broker.username=${BROKER_USERNAME}
     -Dmessaging.broker.password=${BROKER_PASSWORD}
     {{ .Values.environment.ALFRESCO_OPTS | default "" }}
--- a/charts/alfresco-repository/templates/configmap.yaml
+++ b/charts/alfresco-repository/templates/configmap.yaml
@@ -21,12 +21,12 @@ data:
     -Dmetadata-keystore.metadata.algorithm=DESede
     -Dmetadata-keystore.password=mp6yc0UD9e
     -Dmetadata-keystore.metadata.password=oKIWzVdEdA
-    -Dmessaging.broker.url=$(BROKER_URL)
-    -Dmessaging.broker.username=${BROKER_USERNAME}
-    -Dmessaging.broker.password=${BROKER_PASSWORD}
     {{ .Values.environment.ALFRESCO_OPTS | default "" }}
   CATALINA_OPTS: >-
     $ALFRESCO_OPTS
+    -Dmessaging.broker.url="$BROKER_URL"
+    -Dmessaging.broker.username="$BROKER_USERNAME"
+    -Dmessaging.broker.password="$BROKER_PASSWORD"
     {{ .Values.environment.CATALINA_OPTS | default "" }}
   JAVA_OPTS: >-
     {{ .Values.environment.JAVA_OPTS | default "" }}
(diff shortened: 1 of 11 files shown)


charts/alfresco-repository/templates/configmap.yaml
```diff
@@ -2,12 +2,12 @@ apiVersion: v1
...
   name: >-
-    {{- $alfoptsCtx := dict "Values" (dict "nameOverride" "alfresco-opts") "Chart" .Chart "Release" .Release }}
-    {{ template "alfresco-repository.fullname" $alfoptsCtx }}
+    {{- $optsCtx := dict "Values" (dict "nameOverride" "alfresco-opts") "Chart" .Chart "Release" .Release }}
+    {{ template "alfresco-repository.fullname" $optsCtx }}
   labels:
     {{- include "alfresco-repository.labels" . | nindent 4 }}
 data:
-  ALFRESCO_OPTS: >-
+  CATALINA_OPTS: >-
```
gionn, line 10: remember to document the breaking change in acs-deployment

charts/alfresco-repository/values.yaml
```diff
@@ -46,14 +46,37 @@ configuration:
...
-  # -- Alfresco java system properties. These properties must be provided as a
-  # string following the pattern "-Dproperty=value". They override the content
-  # of the global properties file but you should prefer using an existing
-  # configuration.repository.existingConfigMap.
-  ALFRESCO_OPTS: null
-  # -- Apache Tomcat command line options
+  # -- Java or Tomcat system properties. These properties must be provided as a
+  # single string following the pattern "-Dproperty=value -Dmoreprop=morevalue".
+  # They override the content of the global properties file but you should
+  # prefer using an existing configuration.repository.existingConfigMap.
```
gionn, line 79:
```suggestion
  # prefer providing configuration.repository.existingConfigMap
```

charts/alfresco-repository/values.yaml
```diff
@@ -46,14 +46,37 @@ configuration:
...
+    # -- Existing secret and their keys where to find the message broker
+    # username & password.
+    existingSecret:
+      # -- Name of a pre-existing secret containing database credentials
+      name: null
+      keys:
+        # -- Key within the secret holding the database username
+        username: BROKER_USERNAME
+        # -- Key within the secret holding the database password
+        password: BROKER_PASSWORD
```
gionn, line 65:
```suggestion
      # -- Name of a pre-existing secret containing message broker credentials
      name: null
      keys:
        # -- Key within the secret holding the message broker username
        username: BROKER_USERNAME
        # -- Key within the secret holding the message broker password
        password: BROKER_PASSWORD
```

charts/alfresco-repository/values.yaml
```diff
@@ -46,14 +46,37 @@ configuration:
...
     existingSecret: null
+  messageBroker:
+    # -- Message Broker URL
+    url: null
+    # -- Username to authenticate to the message broker
+    username: null
+    # -- Password to authenticate to the message broker
+    password: null
+    # -- Existing secret and their keys where to find the message broker
+    # username & password.
```
gionn, line 57:
```suggestion
```

pretty much the same of what's documented below - and don't generate the ugly object entry in helm docs

charts/alfresco-repository/templates/secret-message-broker.yaml
```diff
@@ -0,0 +1,13 @@
+{{- if and (not .Values.configuration.messageBroker.existingSecret.name) .Values.configuration.messageBroker.username .Values.configuration.messageBroker.password }}
```
gionn, line 1: may not be of high impact, but the upstream activemq doesn't have any credentials enabled by default (or at least that what I remember), maybe we are too strict here? in the end generating a secret with empty credentials is still valid

gionn (changes requested):

charts/alfresco-repository/templates/secret-message-broker.yaml
alxgomz: Not generating the secret at all will result in the same outcome: empty variables si that will work with unauthenticated activemq.

charts/alfresco-repository/values.yaml
alxgomz: it slipped through

charts/alfresco-repository/templates/secret-message-broker.yaml
gionn: true, and I see now the `optional: true` on the ref so it's cool

gionn (approved):

PR #90, opened 2023-08-03
OPSEXP-1862: misc improvements

alxgomz:
Ref: OPSEXP-1862

### DIFF
--- a/charts/alfresco-repository/templates/configmap.yaml
+++ b/charts/alfresco-repository/templates/configmap.yaml
@@ -9,10 +9,10 @@ metadata:
 data:
   CATALINA_OPTS: >-
     -Ddeployment.method=HELM_CHART
-    -Ddb.url=jdbc:{{ .Values.db.url }}
+    -Ddb.url=jdbc:{{ .Values.configuration.db.url }}
     -Ddb.username=${DATABASE_USERNAME}
     -Ddb.password=${DATABASE_PASSWORD}
-    -Ddb.driver={{ include "alfresco-repository.db.driver" .Values.db }}
+    -Ddb.driver={{ include "alfresco-repository.db.driver" .Values.configuration.db }}
     -Dencryption.keystore.type=JCEKS
     -Dencryption.cipherAlgorithm=DESede/CBC/PKCS5Padding
     -Dencryption.keyAlgorithm=DESede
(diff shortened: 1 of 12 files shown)


charts/alfresco-repository/values.yaml
```diff
@@ -16,27 +16,8 @@ strategy:
...
-  # password.
-  existingSecret:
-    # -- Name of a pre-existing secret containing database credentials
-    name: null
-    keys:
-      # -- Key within the secret holding the database username
-      username: DATABASE_USERNAME
-      # -- Key within the secret holding the database password
-      password: DATABASE_PASSWORD
+# -- How long to wait for tomcat to complete shutdown before kiling it
```
gionn, line 19:
```suggestion
# -- How long to wait for tomcat to complete shutdown before killing it
```

charts/alfresco-repository/templates/configmap.yaml
```diff
@@ -9,10 +9,10 @@ metadata:
 data:
   CATALINA_OPTS: >-
     -Ddeployment.method=HELM_CHART
-    -Ddb.url=jdbc:{{ .Values.db.url }}
+    -Ddb.url=jdbc:{{ .Values.configuration.db.url }}
     -Ddb.username=${DATABASE_USERNAME}
     -Ddb.password=${DATABASE_PASSWORD}
-    -Ddb.driver={{ include "alfresco-repository.db.driver" .Values.db }}
+    -Ddb.driver={{ include "alfresco-repository.db.driver" .Values.configuration.db }}
```
gionn, line 15:
```suggestion
    -Ddb.driver={{ include "alfresco-repository.db.driver" .Values.configuration.db.driver }}
```

add a test

gionn (changes requested):

charts/alfresco-repository/templates/configmap.yaml
alxgomz: Here I'm passing the context to the named template so I don't want to pass the driver.
There is already a test for each database type in `tests/config_test.yaml`
https://github.com/Alfresco/alfresco-helm-charts/blob/357447463b465ffd1144f0ad16a7ff65256fdb0a/charts/alfresco-repository/tests/configmap_test.yaml#L49

gionn: oh yeah, looked like a mistype before reading the named template 😀

gionn (approved):

PR #91, opened 2023-08-04
OPSEXP-1862: move named tpl in alfresco-common

alxgomz:
Ref: OPSEXP-18962

Ensure backward compatibility.
Tested backward compat by locally running helm template rendition and unittest for acs-deployment 

Created OPSEXP-2227 OPSEXP-2228 to deprecate and remove old names

### DIFF
--- a/charts/alfresco-common/Chart.yaml
+++ b/charts/alfresco-common/Chart.yaml
@@ -5,7 +5,7 @@ description: |
   A helper subchart to avoid duplication in alfresco charts and set common
   external dependencies
 type: library
-version: 2.1.0-alpha.2
+version: 2.1.0-alpha.3
 dependencies:
   - name: common
     repository: >-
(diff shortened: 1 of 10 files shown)


gionn:
> Ensure backward compatibility.

is it really needed? all the charts pin a version and this chart is not really meant to be used outside of our charts, so... if we need to break an interface is not a big problem

alxgomz:
> > Ensure backward compatibility.
> 
> is it really needed? all the charts pin a version and this chart is not really meant to be used outside of our charts, so... if we need to break an interface is not a big problem

I think so yes. `alfresco-common` is used in every chart and helm i s really bad with having the same dependency in different charts. So I'm afraid even though each chart have its own version of common the one from the parent will take over and we might miss templates although the consuming charts has the right dependency version

gionn (approved):

PR #102, opened 2023-09-01
OPSEXP-1908 Keep appVersion in sync in every chart

gionn:
Ref: OPSEXP-1908

requires https://github.com/Alfresco/alfresco-build-tools/pull/315


.pre-commit-config.yaml
```diff
@@ -40,3 +40,11 @@ repos:
         args:
           - --exclude-charts=alfresco-common,alfresco-sync-service
           - --values-dir=ci
+  - repo: local
+    hooks:
+    -   id: update-charts-yml
+        name: update charts.yml file
+        language: docker_image
+        entry: ghcr.io/alfresco/build-tools-gen-index-yml:OPSEXP-1908@sha256:3c3c6e9bb3df80be1f63bb972588685b5e297faeb7dc56324da5ad24da81294b
```
gionn, line 48: update to master tag once https://github.com/Alfresco/alfresco-build-tools/pull/315 is merged

.pre-commit-config.yaml
```diff
@@ -40,3 +40,11 @@ repos:
         args:
           - --exclude-charts=alfresco-common,alfresco-sync-service
           - --values-dir=ci
+  - repo: local
+    hooks:
+    -   id: update-charts-yml
+        name: update charts.yml file
+        language: docker_image
+        entry: ghcr.io/alfresco/build-tools-gen-index-yml:master@sha256:8e077a69e29aff588221c02c4727525ca21c7912569c40122179086b8477f606
+        files: ^charts/[^/]+/Chart.yaml
```
alxgomz, line 49: seems useless if `pass_filenames` is false

updatecli.d/appVersion.tpl
```diff
@@ -0,0 +1,24 @@
...
+    name: {{ $id }} tag
+    kind: yaml
+    spec:
+      file: charts/{{ $id }}/values.yaml
+      key: {{ if and ($opt) (index $opt "tagKey") }}{{ $opt.tagKey }}{{ else }}$.image.tag{{ end }}
+  {{- end }}
+
+targets:
+  {{- range $id, $opt := .charts }}
+  {{- if and ($opt) (index $opt "noAppVersion") }}{{ continue }}{{ end }}
```
alxgomz, line 16: same

alxgomz (changes requested):

.github/workflows/updatecli.yaml
```diff
@@ -94,14 +94,30 @@ jobs:
...
+        uses: docker/login-action@v2
+        with:
+          registry: ghcr.io
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Regenerate charts.yml if necessary
+        uses: Alfresco/alfresco-build-tools/.github/actions/pre-commit@v3.1.0
+        with:
+          pre-commit-args: update-charts-yml || true
```
alxgomz, line 115: XD
Can't we deal with that in the docker image?

.pre-commit-config.yaml
gionn: it is not, `files` is what triggers the pre-commit to run, while `pass_filenames` is if the changed files needs to be provided as arguments to the invoked `entry`

.github/workflows/updatecli.yaml
gionn: well now that I look it again I should probably have used `continue-on-error` xD

updatecli.d/appVersion.tpl
```diff
@@ -0,0 +1,24 @@
+name: Charts appVersion update
+
+sources:
+  {{- range $id, $opt := .charts }}
+  {{- if and ($opt) (index $opt "noAppVersion") }}{{ continue }}{{ end }}
```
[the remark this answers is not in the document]
gionn: need to have a quick chat about this because the final version of the script is not really helm-aware by choice, see https://github.com/Alfresco/alfresco-build-tools/blob/master/scripts/gen-index-yml/gen_index_yml.py

.pre-commit-config.yaml
alxgomz: Then I would expect the values files to be part of the triggers as well. Am I misunderstanding how this should work?

gionn: it's the simplest way I've found to make sure the hook is run whenever a new chart could be potentially been added (git doesn't store directories so a diff can only be applied to a file - Chart.yaml is a file always present in an helm charts - the script is idempotent and fast to run so should not be a problem to run it when it's not really needed)

PR #105, opened 2023-09-05
OPSEXP-2242: remove usage of lookup in alfresco-repository

alxgomz:
Ref: OPSEXP-2242

Reverting as unusable when called from a parent chart


charts/alfresco-repository/templates/_helpers-database.tpl
```diff
@@ -4,7 +4,7 @@ We're just manipulating the string URl to make it parseable by urlParse.
 It MUST NOT be used directly.
 */}}
 {{- define "alfresco-repository.jdbc.parser" -}}
-{{- $jdbc_url := required "Alfresco repository needs a database to start. Please provide a valid URL in db.url value" . }}
+{{- $jdbc_url := required "Alfresco repository needs a database to start. Please provide a valid URL in configuration.db.url value" . }}
```
gionn, line 7: this error message is not too specific for what this template is doing? I don't expect any values referenced here

if a more specific message is necessary then it needs to be moved before calling this template?

charts/alfresco-repository/templates/_helpers-search.tpl
```diff
@@ -15,19 +15,14 @@ Usage: include "alfresco-repository.search.flavor.valid" "FLAVOR"
 {{/*
 Check whether a Solr shared secret was provided
```
gionn, line 16: check because it doesn't seems to reflect the current implementation

charts/alfresco-repository/tests/search_test.yaml
```diff
@@ -23,69 +37,99 @@ tests:
...
         search:
           << : *solrValues
           solr-secret: null
-    asserts:
-      - matchRegex:
-          path: data.CATALINA_OPTS
-          pattern: |-
-            (^\s*|[^\s]\s+)-Dsolr\.secureComms=none($|\s)
-        template: configmap.yaml
+    asserts: !merge-maps
```
gionn, line 74: what is this?

charts/alfresco-repository/templates/_helpers-database.tpl
alxgomz: wait... actually this template should not even exist in this chart anymore. It's part of `alfresco-common` chart.
Will remove these

charts/alfresco-repository/templates/_helpers-search.tpl
alxgomz: yes it is... the description talks about shared secret as in general authentication concept, not aé kubernetes secrets shared by different charts

charts/alfresco-repository/tests/search_test.yaml
alxgomz: A yaml tag to merge array items (https://github.com/yaml/yaml/issues/48#issuecomment-974998500)

gionn: and it works in our current setup?

alxgomz: Not sure what you mean by that but yes the unit tests pass

charts/alfresco-repository/values.yaml
```diff
@@ -233,3 +239,7 @@ global:
   alfrescoRegistryPullSecrets: quay-registry-secret
   # -- a fallback for .Values.known_urls that can be shared between charts
   known_urls:
+
+# @ignore
```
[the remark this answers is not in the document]
alxgomz: I think it would just be "doc pollution" as no one is ever supposed to use that. If we end up introducing something valuable to users as tags then we would document it

charts/alfresco-repository/templates/_helpers-database.tpl
```diff
@@ -1,105 +1,45 @@
...
-{{- $my_rdbms := dict "name" "mysql" "driver" "com.mysql.jdbc.Driver" }}
-{{- $maria_rdbms := dict "name" "mariadb" "driver" "org.mariadb.jdbc.Driver" }}
-{{- $ora_rdbms := dict "name" "oracle" "driver" "oracle.jdbc.OracleDriver" }}
-{{- $ms_rdbms := dict "name" "sqlserver" "driver" "com.microsoft.sqlserver.jdbc.SQLServerDriver" }}
-{{- range $rdbms := list $pg_rdbms $my_rdbms $maria_rdbms $ora_rdbms $ms_rdbms }}
-{{- eq $rdbms.name $ | ternary $rdbms.driver "" }}
+{{- define "alfresco-repository.db.env" -}}
+{{- $dbcmCtx := dict "Values" (dict "nameOverride" "alfresco-database") "Chart" .Chart "Release" .Release }}
+{{- with .Values.configuration.db }}
+{{- $dbcm := coalesce (tpl .existingConfigMap.name $) (include "alfresco-repository.fullname" $dbcmCtx) }}
```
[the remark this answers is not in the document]
alxgomz: yes so the parent chart can have configmap names that are linked to specific releases

charts/alfresco-repository/templates/configmap-database.yaml
```diff
@@ -0,0 +1,12 @@
...
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: >-
+    {{- $ctx := dict "Values" (dict "nameOverride" "alfresco-database") "Chart" .Chart "Release" .Release }}
+    {{ template "alfresco-repository.fullname" $ctx }}
+  labels:
+    {{- include "alfresco-repository.labels" . | nindent 4 }}
+data:
+  {{ template "alfresco-repository.db.cm" .Values.configuration.db }}
```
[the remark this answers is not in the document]
alxgomz: mostly for readability getting rid of the local variables statements ($dbctx; $dbcm, etc...) and make them part of the helpers

charts/alfresco-repository/templates/_helpers-database.tpl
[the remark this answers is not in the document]
gionn: I am not really excited of using tpl in general because it make things hard to understand when they break (on top of the ugly errors helm already provides out of the box)

what's the problem without it? the user has to provide a different values in case want to do install the same stuff multiple times in the same namespace?

charts/alfresco-repository/values.yaml
```diff
@@ -66,15 +68,15 @@ configuration:
...
+      name: ""
       keys:
         # -- Key within the secret holding the message broker username
         username: BROKER_USERNAME
         # -- Key within the secret holding the message broker password
         password: BROKER_PASSWORD
     existingConfigMap:
       # -- Name of a pre-existing configmap containing the meesage broker URL
-      name: null
+      name: ""
```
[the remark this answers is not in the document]
gionn: well if it's just for that it's better to handle that specific case there and do not pollute values with empty strings instead of plain nulls

charts/alfresco-repository/values.yaml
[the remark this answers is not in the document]
gionn: users are not supposed to use that but I see a value telling them explicitly "this is just for testing the chart - don't use it"

charts/alfresco-repository/templates/_helpers-database.tpl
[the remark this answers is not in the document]
alxgomz: reverted the whole tpl thing for now

gionn (approved):

PR #106, opened 2023-09-06
OPSEXP-2177 Generic database support for search enterprise

gionn:
Ref: OPSEXP-2177

### DIFF
--- a/charts/alfresco-search-enterprise/templates/_helpers-elasticsearch.tpl
+++ b/charts/alfresco-search-enterprise/templates/_helpers-elasticsearch.tpl
@@ -1,5 +1,5 @@
 {{- define "alfresco-search-enterprise.searchIndexExistingSecretName" -}}
-{{ $.Values.searchIndex.existingSecretName | default $.Values.global.elasticsearch.existingSecretName | default (printf "%s-elasticsearch-secret" (include "alfresco-search-enterprise.fullName" $)) }}
+{{ $.Values.searchIndex.existingSecretName | default $.Values.global.elasticsearch.existingSecretName | default (printf "%s-elasticsearch-secret" (include "alfresco-search-enterprise.fullname" $)) }}
 {{- end -}}
 
 {{- define "alfresco-search-enterprise.config.spring" -}}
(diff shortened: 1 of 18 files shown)


charts/alfresco-search-enterprise/templates/_helpers-elasticsearch.tpl
```diff
@@ -1,5 +1,5 @@
 {{- define "alfresco-search-enterprise.searchIndexExistingSecretName" -}}
-{{ $.Values.searchIndex.existingSecretName | default $.Values.global.elasticsearch.existingSecretName | default (printf "%s-elasticsearch-secret" (include "alfresco-search-enterprise.fullName" $)) }}
+{{ $.Values.searchIndex.existingSecretName | default $.Values.global.elasticsearch.existingSecretName | default (printf "%s-elasticsearch-secret" (include "alfresco-search-enterprise.fullname" $)) }}
```
alxgomz, line 2: chained `default` looks ugly

```suggestion
{{ coalesce .Values.searchIndex.existingSecretName .Values.global.elasticsearch.existingSecretName (printf "%s-elasticsearch-secret" (include "alfresco-search-enterprise.fullname" .)) }}
```

charts/alfresco-search-enterprise/templates/configmap-database.yaml
```diff
@@ -0,0 +1,12 @@
...
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: >-
+    {{- $ctx := dict "Values" (dict "nameOverride" "alfresco-database") "Chart" .Chart "Release" .Release }}
+    {{ template "alfresco-search-enterprise.fullname" $ctx }}
+  labels:
+    {{- include "alfresco-search-enterprise.labels" . | nindent 4 }}
+data:
+  {{ .Values.reindexing.db.existingConfigMap.keys.url }}: {{ required ".reindexing.db.url is mandatory when not using existingConfigMap" .Values.reindexing.db.url }}
```
alxgomz, line 11: `.Values.reindexing.db.existingConfigMap.keys.url` really is meant to point to custom cm.
The default value happens to be the same as the one we use in the "embedded" cm but that a "coincidence".

I'm not sure we should do that playing with `.existingConfigMap.keys.*` should not impact the eembeded cm (IMHO)

gionn: true, but it ensure consistency with https://github.com/Alfresco/alfresco-helm-charts/pull/106/files#diff-1f0dd16631d0a9a58373d69ab6cdf59ffcce7fb176c4fcedb37b2526e09bec0aR49

which would break deployment if keys.url is set to something else but embedded cm still enabled

alxgomz: if you provide a key but not the configmap then it's an obvious config error and expected to break in my point of view

gionn: however it will increase the complexity inside the deployment where we need to make sure to use the correct key when using existingConfigMap or not, so it's actually less code and more robust if we stick to this approach.

alxgomz: I still don't agree as the additionnal "robustness" only covers use cases which should break imho. But for now let's move on as we anyway have a standardization work to do

alxgomz (approved):

PR #107, opened 2023-09-07
OPSEXP-1908 Enable appVersion bump from uber manifest

gionn:
Ref: OPSEXP-1908

Requires https://github.com/Alfresco/alfresco-updatecli/pull/8

### DIFF
--- a/.github/workflows/updatecli.yaml
+++ b/.github/workflows/updatecli.yaml
@@ -88,7 +88,7 @@ jobs:
         uses: actions/checkout@v3
         with:
           repository: alfresco/alfresco-updatecli
-          ref: master
+          ref: OPSEXP-1908-appVersion
           path: alfresco-updatecli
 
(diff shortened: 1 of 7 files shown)


.github/workflows/updatecli.yaml
```diff
@@ -88,7 +88,7 @@ jobs:
         uses: actions/checkout@v3
         with:
           repository: alfresco/alfresco-updatecli
-          ref: master
+          ref: OPSEXP-1908-appVersion
```
gionn, line 91:
```suggestion
          ref: master
```

alxgomz (approved):
Is there any reason why we're having `7.4.N` as a matrix key here?
Maybe worth fixing at the same time?

gionn:
> Is there any reason why we're having `7.4.N` as a matrix key here?

we want to track latest GA which is 7.4 - so it's good as it is now

PR #109, opened 2023-09-15
OPSEXP-2242: add support for existing secret in search service

alxgomz:
Ref: OPSEXP-2242

For a integration in acs-deployment 7.0.0


charts/alfresco-search-service/ci/default-values.yaml
```diff
@@ -2,6 +2,5 @@
 alfresco-insight-zeppelin:
   nameOverride: zeppelin
   enabled: true
-global:
-  tracking:
-    sharedsecret: dummy
+repository:
+  shared-secret: dummy
```
gionn, line 6: use camelCase as per usual k8s standards

charts/alfresco-search-service/templates/deployment.yaml
```diff
@@ -89,26 +93,32 @@ spec:
...
-              {{- end }}
+            exec:
+              command:
+                - sh
+                - -c
+                - >-
+                  [
+                  $(
+                  curl -sH "X-Alfresco-Search-Secret: $SOLR_ALFRESCO_SECURECOMMS_SECRET" http://localhost:{{ template "alfresco-search.containerPort" . }}/solr/alfresco/admin/ping
+                  | xmllint --xpath '/response/str[@name="status"]/text()' -
```
gionn, line 104: so the http response doesn't reflect the service status?

charts/alfresco-search-service/values.yaml
```diff
@@ -64,12 +64,16 @@ ingress:
 repository:
   # -- Alfresco repository URL
   url: http://alfresco-search-service/solr
+  # -- repository seucurity level to use when tracking the repo ('none' or 'secret')
```
gionn, line 67:
```suggestion
  # -- repository security level to use when tracking the repo ('none' or 'secret')
```

charts/alfresco-search-service/values.yaml
```diff
@@ -64,12 +64,16 @@ ingress:
...
+  securecomms: secret
```
gionn, line 68: I wonder if we really want to mimic the app behavior here or at least provide a more meaningful name for this key, or we could also provide a boolean instead which internally map to `none` or `secret`?

gionn (changes requested):

charts/alfresco-search-service/templates/deployment.yaml
alxgomz: not in case of 403. The exit code would still be 0 even though the server denied access

charts/alfresco-search-service/values.yaml
alxgomz: We can rule out the boolean as there is actually 3 possible values (`none`, `secret` & `https`).
As for changing the name I'm not sure that's a good idea as the meaning differs based on the search engine. `https` for solr means mTLS while for elasticsearch it can be simply encrypted http. Sticking to what people already know might be less confusing for them.

gionn (approved):

PR #115, opened 2023-09-21
OPSEXP-2286: allow passing the ATS location

alxgomz:
Ref: OPSEXP-2286

Previous the ATS location was only computed using an old chart and only worked if the ats router and sfs were using specific name prefixes (alfresco-cs & alfresco)

That PR allow passing the ATS details in values or configmap, making either of both mandatory.


charts/alfresco-search-enterprise/values.yaml
```diff
@@ -88,14 +88,24 @@ reindexing:
         limits:
           cpu: "0.25"
           memory: "10Mi"
+ats:
+  # -- URL of the alfresco transform (trouter or tengine-aio)
+  transform_url: null
+  # -- URL of the alfresco shared filestore (trouter or tengine-aio)
```
gionn, line 94: what `(trouter or tengine-aio)` means in this comment?

charts/alfresco-search-enterprise/values.yaml
```diff
@@ -88,14 +88,24 @@ reindexing:
...
+    keys:
+      # -- Key within the configmap holding the URL of the alfresco transform
+      transform_url: ATS_URL
+      # -- Key within the configmap holding the URL of the alfresco shared filestore
+      sfs_url: SFS_URL
 elasticsearch:
   # -- Enable embedded elasticsearch - useful when using this chart in standalone
   enabled: false
   replicas: 1
   clusterHealthCheckParams: "wait_for_status=yellow&timeout=1s"
```
gionn, line 108: do you want to move also this to the bottom if the intention is to have all values related to testing at the bottom?

charts/alfresco-search-enterprise/templates/_helpers-elasticsearch.tpl
```diff
@@ -1,14 +1,14 @@
 {{- define "alfresco-search-enterprise.searchIndexExistingSecretName" -}}
-{{ coalesce .Values.searchIndex.existingSecretName .Values.global.elasticsearch.existingSecretName (printf "%s-elasticsearch-secret" (include "alfresco-search-enterprise.fullname" .)) }}
+{{- coalesce .Values.searchIndex.existingSecretName .Values.global.elasticsearch.existingSecretName (printf "%s-elasticsearch-secret" (include "alfresco-search-enterprise.fullname" .)) }}
 {{- end -}}
 
 {{- define "alfresco-search-enterprise.config.spring" -}}
-{{- if .Values.elasticsearch.enabled }}
+{{- if .Values.elasticsearch.enabled -}}
```
gionn, line 6: is the closing `-` really necessary?

gionn (approved):

charts/alfresco-search-enterprise/values.yaml
alxgomz: it means I'm a copy/paste maniac... will fix thx.

charts/alfresco-search-enterprise/templates/_helpers-elasticsearch.tpl
alxgomz: that's cosmetic tbh 💅 ut I would argue that a named template should never start with a carriage return.

PR #119, opened 2023-09-26
OPSEXP-2242: make repository secrets optional

alxgomz:
Ref: OPSEXP-2242: Needed in order to pass sensitive data (liked to a specific feature or config, e.g. smtp password) from a parent chart without checking whether the eature is enabled in subchart.
Open to discussion.

### DIFF
--- a/charts/alfresco-repository/Chart.yaml
+++ b/charts/alfresco-repository/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: alfresco-repository
 description: Alfresco content repository Helm chart
 type: application
-version: 0.1.0-alpha.9
+version: 0.1.0-alpha.10
 appVersion: 23.1.0-A21
 dependencies:
   - name: alfresco-common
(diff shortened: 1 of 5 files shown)


gionn (commented):
at first sight I don't get the point here. If I am passing existingSecrets I need to ensure that they really exists, no matter when they get created, the repository pod has to wait for them, otherwise it should just wait for them (in PodInitialize state) as it is now

alxgomz:
> at first sight I don't get the point here. If I am passing existingSecrets I need to ensure that they really exists, no matter when they get created, the repository pod has to wait for them, otherwise it should just wait for them (in PodInitialize state) as it is now
The problem comes when the parent chart offers pass that secret as an option. let say on the parent chart, using `feature.enabled ` enables a feature which requires a secret. When the feature is enabled the parent needs to creates the secret and pass it to the subchart as an `existingSecrets` entry. But if that feature is not enabled, then the secret is not created and the pods in the subchart will never start unless the `existingSecret` is either:
* set to null manualy
* rendered dynamically by the subchart (which requires both `tpl()` & some global scoped values)
* made optional

gionn:
understood, let's try it, but there is a risk that something will not work as expected if the repo pod get initialized before the secret actually created by the parent chart.

gionn (approved):

PR #120, opened 2023-09-26
OPSEXP-2242: propertify secrets

alxgomz:
Allow passing sensitive data to alfresco-repository as secrets and leverage those as properties read form the env var


charts/alfresco-repository/values.yaml
```diff
@@ -25,10 +25,12 @@ configuration:
     # with actual Alfresco repository properties
     existingConfigMap: null
     # -- A list of secrets to make available to the repository as env vars.
+    # If the secrets neds to be used by the repo as a property on can use the
```
gionn, line 28:
```suggestion
    # If the secrets needs to be used by the repo as a property one can use the
```

charts/alfresco-repository/values.yaml
```diff
@@ -25,10 +25,12 @@ configuration:
...
+    # following purpose syntax: 'property:name.of.the.prop'
     # This list can contain special secrets marked with predifined `purpose`:
     # `acs-license` to pass license as a secret or subsystems:*:* to configure
     # an Alfresco subsystem. See [Configuring Alfresco
-    # Subsystem](./docs/subsystems.md) for more details.
+    # Subsystem](./docs/subsystems.mq) for more details.
```
gionn, line 33: revert

charts/alfresco-repository/values.yaml
```diff
@@ -25,10 +25,12 @@ configuration:
     # with actual Alfresco repository properties
     existingConfigMap: null
     # -- A list of secrets to make available to the repository as env vars.
+    # If the secrets neds to be used by the repo as a property on can use the
+    # following purpose syntax: 'property:name.of.the.prop'
```
gionn, line 29:
```suggestion
    # following syntax: 'property:name.of.the.prop'
```

charts/alfresco-repository/templates/configmap-repository.yaml
```diff
@@ -27,6 +27,12 @@ data:
     -Dmessaging.broker.password="$BROKER_PASSWORD"
     -Dindex.subsystem.name="$SEARCH_FLAVOR"
     {{- include "alfresco-repository.search.config" $ | indent 2 }}
+    {{- range .repository.existingSecrets }}
+      {{- if and .key (hasPrefix "property:" .purpose) }}
+        {{- $property := trimPrefix "property:" .purpose }}
+        {{- printf "-D%s=\"$%s\"" $property .key | nindent 4 }}
+      {{- end }}
+    {{- end }}
```
gionn, line 35: not sure having one secret for each property scale well but if that's for a limited number of use case ok

gionn (changes requested):

charts/alfresco-repository/templates/configmap-repository.yaml
alxgomz: I see it as a general way to pass sensitive properties while getting the value from env vars and avoid storing it in a file.
I don't think there would be that many. Is there a particular concern of why that would not scale?

PR #121, opened 2023-09-27
OPSEXP-2242: add custom keystores support

alxgomz:
Ref: OPSEXP-2242

### DIFF
--- a/charts/alfresco-repository/docs/keystores.md
+++ b/charts/alfresco-repository/docs/keystores.md
@@ -1,0 +1,83 @@
+# Using custom keystore
+
+Alfresco uses a set of keystores for different purposes. Among others thing
+keystores can be used for metadata encryption or traffic encryption.
+
+The alfresco-content-repository image comes with a keystore which contains a
+single security key. That key is primarily used for metadata encryption and it
+required when the repository starts up. In production environment it is
+recommended to use a custom keystore rather than the one present in the image.
+
+One may also need to have additional keystores deployed, for instance to allow
+using TLS encryption for a service the repository provides using an x509
+certificate. We describe below how to do both.
+
+## The metadata keystore
+
+This keystore is used for property encryption and needs to remain identical
+throughout the life of the repository. Losing this keystore or its passwords
+will result in a repository that cannot startup so it is absolutely crucial to
+store these assets in a version control system like git (both the keystore and
+its password).
+
+First create a local keystore. It is recommended to use p12 containers nowadays
+and leverage AES algorithm:
+
+```bash
+keytool -genseckey \
+  -keystore /path/to/keystore.p12
+  -storetype PKCS12 \
+  -storepass secret -keyalg AES \
+  -keysize 256 \
+  -alias metadata
+```
+
+Now we need to create a secret with from that keystore file and its password in
+the namespace where Alfresco will be deployed:
+
+```bash
+kubectl create secret generic repository-keystores \
+  --from-file=/path/to/keystore.p12 \
+  --from-literal "ENC_METADATA_STOREPASS=secret" \
+  --from-literal "ENC_METADATA_KEYPASS=secret"
+```
+
+> Note, unlike the older JCEKS format, PKCS12 do not allow having different
+> passwords for the keystore and the key. so both variables need to be
+> populated with the same value (the same one that was used whn generating the
+> keystore using keytool (here "secret").
+
+Once the secret is created in target namespace we need to let the chart know it
+needs to mount it and provide necessary options for the repository to access
+that keystore. To do so use the properties below:
+
+```yaml
+---
+alfresco-repository:
+  configuration:
+    repository:
+      existingSecrets:
+        - name: repository-keystores
+          key: ENC_METADATA_STOREPASS
+        - name: repository-keystores
+          key: ENC_METADATA_KEYPASS
+  environment:
+    JAVA_OPTS: >-
+      -Dencryption.keystore.type=PKCS12
+      -Dencryption.cipherAlgorithm=AES/CBC/PKCS5Padding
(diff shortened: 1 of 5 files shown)


charts/alfresco-repository/docs/keystores.md
```diff
@@ -0,0 +1,83 @@
...
+using TLS encryption for a service the repository provides using an x509
+certificate. We describe below how to do both.
+
+## The metadata keystore
+
+This keystore is used for property encryption and needs to remain identical
+throughout the life of the repository. Losing this keystore or its passwords
+will result in a repository that cannot startup so it is absolutely crucial to
+store these assets in a version control system like git (both the keystore and
+its password).
```
gionn, line 21: I would not suggest git to store credentials, just say that is important to properly back it up together with data otherwise it will be unusable.

charts/alfresco-repository/templates/deployment.yaml
```diff
@@ -68,6 +68,10 @@ spec:
             {{- $searchsecret = coalesce .search.existingSecret.name (include "alfresco-repository.fullname" $searchsecretCtx) }}
           {{- end }}
           env:
+            - name: ENC_METADATA_STOREPASS
+              value: mp6yc0UD9e
+            - name: ENC_METADATA_KEYPASS
+              value: oKIWzVdEdA
```
gionn, line 74: these get overridden by the latter defined when using existingSecret right?

gionn (changes requested):

charts/alfresco-repository/templates/deployment.yaml
alxgomz: indeed. One needs to pass the followig yaml to override these values:

```yaml
  configuration:
    repository:
      existingSecrets:
        - name: repository-keystores
          key: ENC_METADATA_STOREPASS
        - name: repository-keystores
          key: ENC_METADATA_KEYPASS
```

charts/alfresco-repository/docs/keystores.md
```diff
@@ -0,0 +1,83 @@
...
+Alfresco uses a set of keystores for different purposes. Among others thing
+keystores can be used for metadata encryption or traffic encryption.
+
+The alfresco-content-repository image comes with a keystore which contains a
+single security key. That key is primarily used for metadata encryption and it
+required when the repository starts up. In production environment it is
+recommended to use a custom keystore rather than the one present in the image.
+
+One may also need to have additional keystores deployed, for instance to allow
+using TLS encryption for a service the repository provides using an x509
```
aborroy, line 12: TLS >> mTLS

charts/alfresco-repository/templates/configmap-repository.yaml
```diff
@@ -14,26 +14,26 @@ data:
...
-      {{- if and .key (hasPrefix "property:" .purpose) }}
+      {{- if and .key (hasPrefix "property:" (.purpose |  default "")) }}
         {{- $property := trimPrefix "property:" .purpose }}
         {{- printf "-D%s=\"$%s\"" $property .key | nindent 4 }}
       {{- end }}
     {{- end }}
     {{- end }}
     {{ .Values.environment.CATALINA_OPTS | default "" }}
   JAVA_OPTS: >-
+    -Dencryption.keystore.type=JCEKS
```
aborroy, line 33: JCEKS >> PKCS12

charts/alfresco-repository/templates/configmap-repository.yaml
```diff
@@ -14,26 +14,26 @@ data:
...
+    -Dencryption.cipherAlgorithm=DESede/CBC/PKCS5Padding
```
aborroy, line 34: DESede/CBC/PKCS5Padding >> AES/CBC/PKCS5Padding

charts/alfresco-repository/templates/configmap-repository.yaml
```diff
@@ -14,26 +14,26 @@ data:
...
+    -Dencryption.keyAlgorithm=DESede
```
aborroy, line 35: DESede >> AES

charts/alfresco-repository/templates/configmap-repository.yaml
```diff
@@ -14,26 +14,26 @@ data:
...
+    -Dencryption.keystore.location=/usr/local/tomcat/shared/classes/alfresco/extension/keystore/keystore
+    -Dmetadata-keystore.aliases=metadata
+    -Dmetadata-keystore.metadata.algorithm=DESede
```
aborroy, line 38: DESede >> AES

aborroy (changes requested):
Documentation and implementation are out of sync.

alxgomz:
> Documentation and implementation are out of sync.

Actually the lines you're pointing are using the old option because this is what the image ships. So changing them would prevent the repo from starting up. The documentation explains how to generate a custom keystore and pass appropriate values to override these defaults.
Another option would have been to configure the repo to find its keystore in the contentstore volume (writable to the container) and let alfresco generate its keystore on bootstrap using modern options (which are enabled by default AFAICT https://github.com/Alfresco/alfresco-community-repo/blob/eebd110c3422ec9b1920033d100ea6784c713fa7/repository/src/main/resources/alfresco/repository.properties#L708) but I'm reluctant to that as it would be so easy for people to generate a keystore, forget about, lose it and scream

charts/alfresco-repository/docs/keystores.md
alxgomz: I'm not mentioning mTLS here on purpose as it is very likely that the other end TLS session might not have the same support for keystore. For now this is limited to repository

aborroy (approved):
LGTM

gionn (approved):

PR #131, opened 2023-10-18
OPSEXP-2298: provide common templates to build activemq configmap

alxgomz:
Ref: OPSEXP-2298
Needed for search enterprise chart (and future) others to avoid/reduce duplication

### DIFF
--- a/charts/alfresco-common/templates/_helpers-activemq.tpl
+++ b/charts/alfresco-common/templates/_helpers-activemq.tpl
@@ -1,4 +1,28 @@
 {{/*
+Validate ActiveMQ has a failover transport URL
+
+Usage: include "alfresco-common.activemq.url.valid" "URL"
+
+*/}}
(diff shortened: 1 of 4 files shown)


charts/alfresco-common/templates/_helpers-activemq.tpl
```diff
@@ -1,3 +1,27 @@
+{{/*
+Validate ActiveMQ has a failover transport URL
+
+Usage: include "alfresco-common.activemq.url.valid" "URL"
+
+*/}}
+{{- define "alfresco-common.activemq.url.valid" -}}
```
gionn, line 7: try to have a name which reflect the current behaviour - remember that changing the implementation without changing the template name is risky

something like `alfresco-common.activemq.url.withFailover`?

gionn (changes requested):

gionn (approved):

PR #134, opened 2023-10-19
OPSEXP-2298: fixup search ernterprise chart - 3.0.0-alpha.4

alxgomz:
Ref: OPSEXP-2298

looks like the `$(VAR)` notation in `spec.template.spec.containers[].args` does not work in shell and just creates a subshell instead (as it would in a regular shell env).
Tricked by https://kubernetes.io/docs/tasks/inject-data-application/define-environment-variable-container/#using-environment-variables-inside-of-your-config


gionn (approved):
🤕

PR #135, opened 2023-10-19
OPSEXP-2298: add support for generalized existingSecret in search enterprise

alxgomz:
Ref: OPSEXP-298

### DIFF
--- a/charts/alfresco-search-enterprise/templates/_helpers-activemq.tpl
+++ b/charts/alfresco-search-enterprise/templates/_helpers-activemq.tpl
@@ -20,11 +20,31 @@ Usage: include "alfresco-search-enterprise.activemq.cm.env" $
 
 */}}
 {{- define "alfresco-search-enterprise.activemq.cm.env" -}}
-{{- $mqCtx := dict "Values" (dict "nameOverride" (printf "%s-%s" ($.Values.nameOverride | default $.Chart.Name) "mq")) "Chart" $.Chart "Release" $.Release }}
+{{- $mqCtx := dict "Values" (dict "nameOverride" (printf "%s-%s" (.Values.nameOverride | default $.Chart.Name) "mq")) "Chart" .Chart "Release" .Release }}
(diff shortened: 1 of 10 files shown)


charts/alfresco-search-enterprise/ci/default-values.yaml
```diff
@@ -34,3 +25,9 @@ elasticsearch:
     limits:
       cpu: "1000m"
       memory: "1Gi"
+activemq:
+  fullnameOverride: activemq
```
gionn, line 29: why you dropped resources?

I think at least requests.cpu is always helpful to have to avoid failures related to low cpu availability of the runners

charts/alfresco-search-enterprise/templates/_helpers-activemq.tpl
```diff
@@ -4,14 +4,7 @@ Usage: include "alfresco-search-enterprise.activemq.url" $
 
 */}}
 {{- define "alfresco-search-enterprise.activemq.url" -}}
-{{- with .Values.activemq }}
-{{- if .enabled }}
-{{- $mqCtx := dict "Values" (dict "nameOverride" (.nameOverride | default "")) "Chart" (dict "Name" "activemq") "Release" $.Release }}
-{{- printf "failover:(nio://%s-broker:61616)?timeout=3000&jms.useCompression=true" (include "activemq.fullname" $mqCtx) }}
-{{- else }}
 {{- required "Disabling in-cluster ActiveMQ requires passing (at least) messageBroker.url" $.Values.messageBroker.url }}
```
gionn, line 7: probably the message needs to be revisited given technically we are not managing any more in-cluster activemq here

charts/alfresco-search-enterprise/ci/default-values.yaml
alxgomz: I think I tought Iwas fiddling with the unit test values

gionn (approved):

PR #140, opened 2023-10-30
OPSEXP-2297: fixup & revert optional driverclass

alxgomz:
Ref: OPSEXP-2297

Making the DATABASE_DRIVER var optional didn't work unless url was set b values (reverted).
Siwtched the replacement of DATABASE_HOST & DATABASE_PORT from value templating to env variables within the pod (also didn't play well when param set using existingCm


charts/alfresco-repository/templates/configmap-repository.yaml
```diff
@@ -16,7 +16,7 @@ data:
     -Ddb.url="$DATABASE_URL"
     -Ddb.username=${DATABASE_USERNAME}
     -Ddb.password=${DATABASE_PASSWORD}
-    -Ddb.driver="{{ printf "${DATABASE_DRIVER:-%s}" (include "alfresco-common.db.driver" .db) }}"
+    -Ddb.driver="${DATABASE_DRIVER:-%s}"
```
gionn, line 19: %s?

gionn (changes requested):

gionn (approved):

PR #144, opened 2023-11-06
OPSEXP-2292 Review and align chart activemq with newer principals

slohe1:
OPSEXP-2292


gionn (approved):
adminUser.* is not totally like in other chart but I will avoid to break again the compatibility here given that the feature is still usable (except the customization of key names in the secret)

PR #151, opened 2023-11-08
OPSEXP-2364 Ensure release charts doesn't have pre-release dependencies

gionn:
Ref: OPSEXP-2364

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -18,8 +18,8 @@ repos:
         files: charts/.*\.yaml$
         verbose: true
         args:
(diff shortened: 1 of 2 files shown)


gionn:
> I would have checked that at release time only (so probably not with pre-commit) but otherwise LGTM. Do you want to add more stuff (still in draft)?

this time really forgot because I was in a hurry :D

I put it in pre-commit because the check is fine to be always enforced: if you want to use an alpha dependency then you never want to release it with a non-alpha chart.

alxgomz (approved):

PR #160, opened 2023-12-08
added podannotations for search-services

shazada:
we use podAnnotations in the 'charts' so would like to have these in all chart, but for now atleast in search-services.

### DIFF
--- a/charts/alfresco-search-service/templates/deployment.yaml
+++ b/charts/alfresco-search-service/templates/deployment.yaml
@@ -21,6 +21,9 @@ spec:
     metadata:
       annotations:
         checksum/config: {{ include (print $.Template.BasePath "/config.yaml") . | sha256sum }}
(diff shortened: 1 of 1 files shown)


shazada:
I've just send you an invite so hopefully that  works

shazada:
Lol I can't seem to see the feature.
Screenshot_20231208_194342_Firefox

alxgomz:
It should be present next to the "Comment" button in your PR:

image

Is it not?

shazada:
image

alxgomz:
After digging a bit it looks like we cannot contribute on PR whic are not coming from "user-owned" forks. I guess "Contezza" is a github org, so we'll have to create our own PR to add version bumps and tests.
Thanks for your contrib though!

alxgomz:
superseeded by #161

PR #162, opened 2023-12-11
Fixup and bump pre-commit hooks

gionn:


gionn:
There is a fresh issue on this failure https://github.com/bridgecrewio/checkov/issues/5841

alxgomz (approved):

PR #166, opened 2023-12-13
missing component search on search-services deployment

shazada:
In the pod labels, search is missing the component search.

Which other deployment have, like transformers for all the tansformer pods.

image

image

This time through my personal github account 👍

### DIFF
--- a/charts/alfresco-search-service/templates/deployment.yaml
+++ b/charts/alfresco-search-service/templates/deployment.yaml
@@ -27,6 +27,7 @@ spec:
       labels:
         app: {{ template "alfresco-search-service.fullname" . }}-solr
         release: {{ .Release.Name }}
+        component: search
     spec:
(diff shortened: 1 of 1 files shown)


alxgomz:
Hi @shazada ,

We acknoledge a discrepency in the labels the charts apply to their pods. That's probably due to a not very consistent work when moving to individual charts...
That said the label you would like to (re-) introduce are rather old form and we'd rather move them to the newer recomanded lables as advertised in https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/

I hope that will work for you once all charts are aligned to this convention.

shazada:
Thanks for the update. If I can use the newer for in the network-policy then that is fine. Now were using component search or transform, etc.

gionn:
missing component labels has been provided in all charts with https://github.com/Alfresco/alfresco-helm-charts/pull/213

PR #167, opened 2023-12-18
OPSEXP-2293: align alfresco-ai-transformer chart with latest practices

alxgomz:
Ref: OPSEXP-2293


charts/alfresco-ai-transformer/README.md
```diff
@@ -1,26 +1,32 @@
...
-| aws.comprehendRoleARN | string | `"arn:aws:iam::XXXXXXXXXXXX:role/ComprehendAsyncJobs"` |  |
-| aws.region | string | `"region-name"` |  |
-| aws.s3Bucket | string | `"s3-bucket-name"` |  |
-| aws.secretAccessKey | string | `"XXXXXXXXXXXXXXXXXXXXXXXX"` |  |
+| aws.accessKeyId | string | `nil` | AWS credentials are required as documented at https://docs.alfresco.com/intelligence-services/latest/config/#default-configuration |
+| aws.comprehendRoleARN | string | `nil` |  |
+| aws.existingConfigMap.keys.comprehendRoleARN | string | `"AWS_COMPREHEND_ROLE_ARN"` |  |
+| aws.existingConfigMap.keys.region | string | `"AWS_REGION"` |  |
+| aws.existingConfigMap.keys.s3Bucket | string | `"AWS_S3_BUCKET"` |  |
+| aws.existingConfigMap.name | string | `nil` |  |
```
gionn, line 23: not sure we really need to support this configMap, what's the reasoning behind?

if they are not secrets and nor service URLs or any kind of configuration which can be reused elsewhere...

alxgomz: The reasoning was to provide both values and cm for all config assets.If we chose to exclude some configs from this patern then I would remove both (dedicated) values & cm and simply use generic `.Values.environment`.
What do you think?

gionn: it can be a great idea given that we are not really solving any problem by adding another layer of variables when we can simply reference the same variables as per docs

gionn (commented):
simply use generic .Values.environment

gionn (approved):

charts/alfresco-ai-transformer/README.md
alxgomz: gave it a second thought and disccussed it internally. We concluded we xwwould keep the PR as is as worst case I see is that no one really uses that configmap. But that do not prevents anyone from working the old way

PR #172, opened 2023-12-20
OPSEXP-2293: use ATS common named templates in ai & search ent charts

alxgomz:
Ref: OPSEXP-2293

### DIFF
--- a/charts/alfresco-search-enterprise/templates/_helpers-ats.tpl
+++ b/charts/alfresco-search-enterprise/templates/_helpers-ats.tpl
@@ -6,7 +6,7 @@ Usage: include "alfresco-search-enterprise.transform.fullurl" "URL"
 
 */}}
 {{- define "alfresco-search-enterprise.ats.fullurl" -}}
-  ATS_URL: {{ printf "%s/transform/config" . }}
+  ATS_URL: {{ template "alfresco-common.ats.fullurl" . }}
 {{- end -}}
 
 {{/*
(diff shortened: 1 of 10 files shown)


charts/alfresco-search-enterprise/templates/_helpers-ats.tpl
```diff
@@ -17,7 +17,7 @@ Usage: include "alfresco-search-enterprise.sfs.fullurl" "URL"
 
 */}}
 {{- define "alfresco-search-enterprise.sfs.fullurl" -}}
```
gionn, line 19: I guess you can drop these two named templates and inline them given they are used only in a single place and without any additional wrapping logic

alxgomz: I could but I decided to not do it to avoid a major bump

gionn: why major bump? those named templates are meant to be used internally and we are not changing any behavior (hopefully)

alxgomz: they are used [here](https://github.com/Alfresco/acs-deployment/blob/2c0e32ba03727e4a3f5294d65879edfadf8a42a6/helm/alfresco-content-services/templates/config-infrastructure.yaml#L59)

gionn (approved):

PR #175, opened 2024-01-02
OPSEXP-2301 Bump ATS chart to most recent conventions

gionn:
* add existing secret support for message broker credentials
* add existing configmap support for message broker url
* service account
* tolerations/affinity
* pod annotations/labels
* volumes/volumeMounts

OPSEXP-2301


charts/alfresco-transform-service/templates/_helpers-name.tpl
```diff
@@ -33,6 +33,11 @@
 {{- include "alfresco-transform-service.fullname" $scope }}
 {{- end }}
 
+{{- define "alfresco-transform-service.config-messagebroker.name" -}}
+{{- $scope := (dict "Values" (dict "nameOverride" "messagebroker-configmap" ) "Chart" .Chart "Release" .Release) }}
```
alxgomz, line 37: I think using "component-objecttype" as names is not very good. It's both reduddant (as you normally know what obj type you're querying) and not self explanatory (here it's not giving any about the fact the cm is used by ats)
I would prefer using something like "message-broker-ats" or "alfresco-transform-service-mq"

charts/alfresco-transform-service/templates/_helpers-name.tpl
```diff
@@ -67,3 +72,8 @@
 {{- $scope := (dict "Values" (dict "nameOverride" "transform-router" ) "Chart" .Chart "Release" .Release) }}
 {{- include "alfresco-transform-service.fullname" $scope }}
 {{- end }}
+
+{{- define "alfresco-transform-service.secret-messagebroker.name" -}}
+{{- $scope := (dict "Values" (dict "nameOverride" "messagebroker-secret" ) "Chart" .Chart "Release" .Release) }}
+{{- include "alfresco-transform-service.fullname" $scope }}
+{{- end }}
```
alxgomz, line 79: IMHO cm and secret can/should share the same name

charts/alfresco-transform-service/templates/deployment-filestore.yaml
```diff
@@ -63,6 +67,24 @@ spec:
...
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
+      {{- with .Values.filestore.nodeSelector }}
+      nodeSelector:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
+      {{- with .Values.filestore.affinity }}
+      affinity:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
```
alxgomz, line 85: duplicates the affinity definition in line 33.
Please move it to the default values and replace the deprecated `failure-domain.beta.kubernetes.io/zone` by the newer `topology.kubernetes.io/zone`

charts/alfresco-transform-service/templates/deployment-imagemagick.yaml
```diff
@@ -53,14 +57,32 @@ spec:
...
+      {{- with .Values.imagemagick.volumes }}
+      volumes:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
+      {{- with .Values.imagemagick.nodeSelector }}
+      nodeSelector:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
+      {{- with .Values.imagemagick.affinity }}
+      affinity:
```
alxgomz, line 81: requries refactoring of affinity definition line 33

charts/alfresco-transform-service/templates/deployment-libreoffice.yaml
```diff
@@ -53,14 +57,32 @@ spec:
...
+      {{- with .Values.libreoffice.volumes }}
+      volumes:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
+      {{- with .Values.libreoffice.nodeSelector }}
+      nodeSelector:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
+      {{- with .Values.libreoffice.affinity }}
+      affinity:
```
alxgomz, line 81: same as above

charts/alfresco-transform-service/templates/deployment-pdfrenderer.yaml
```diff
@@ -53,14 +57,32 @@ spec:
...
+      {{- with .Values.pdfrenderer.volumes }}
+      volumes:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
+      {{- with .Values.pdfrenderer.nodeSelector }}
+      nodeSelector:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
+      {{- with .Values.pdfrenderer.affinity }}
+      affinity:
```
alxgomz, line 81: same as above

charts/alfresco-transform-service/templates/deployment-tika.yaml
```diff
@@ -53,14 +57,32 @@ spec:
...
+      {{- with .Values.tika.volumes }}
+      volumes:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
+      {{- with .Values.tika.nodeSelector }}
+      nodeSelector:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
+      {{- with .Values.tika.affinity }}
+      affinity:
```
alxgomz, line 81: same as above

charts/alfresco-transform-service/templates/deployment-transform-router.yaml
```diff
@@ -53,25 +57,41 @@ spec:
...
           resources: {{- toYaml .Values.transformrouter.resources | nindent 12 }}
           {{ include "alfresco-transform-service.probes" .Values.transformrouter | nindent 10 }}
-          {{- if .Values.ai.enabled }}
           volumeMounts:
-          - name: config-volume
-            mountPath: /mnt/routes
+          {{- if .Values.ai.enabled }}
+            - name: config-volume
+              mountPath: /mnt/routes
+          {{- end }}
```
alxgomz, line 72: You can remove that as I think it's better to deal with it from the umbrella chart as a generic extra volumeMount

charts/alfresco-transform-service/templates/deployment-transform-router.yaml
```diff
@@ -53,25 +57,41 @@ spec:
...
-      {{- if .Values.ai.enabled }}
       volumes:
-      - name: config-volume
-        configMap:
-          name: {{ template "alfresco.shortname" . }}-transform-routes-configmap
+        {{- if .Values.ai.enabled }}
+        - name: config-volume
+          configMap:
+            name: {{ template "alfresco.shortname" . }}-transform-routes-configmap
+        {{- end }}
```
alxgomz, line 81: Should be handled with a generic extra volume

charts/alfresco-transform-service/templates/deployment-transform-router.yaml
```diff
@@ -53,25 +57,41 @@ spec:
...
+        {{- with .Values.transformrouter.volumes }}
+        {{- toYaml . | nindent 8 }}
+        {{- end }}
+      {{- with .Values.transformrouter.nodeSelector }}
+      nodeSelector:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
+      {{- with .Values.transformrouter.affinity }}
+      affinity:
```
alxgomz, line 90: needs refactoring of predefined affinity line 33

charts/alfresco-transform-service/values.yaml
```diff
@@ -304,32 +344,46 @@ filestore:
...
+  existingSecret:
+    # -- Alternatively, provide message broker credentials via an existing Secret
+    name: null
+    keys:
+      user: BROKER_USERNAME
+      password: BROKER_PASSWORD
+
 ai:
   enabled: false
   nameOverride: alfresco-ai
```
alxgomz, line 371: to remove

charts/alfresco-transform-service/values.yaml
```diff
@@ -304,32 +344,46 @@ filestore:
...
+  volumes: []
+  volumeMounts: []
+
 messageBroker:
+  # -- External activemq connection url (e.g. failover:(nio://my-broker:61616)?timeout=3000&jms.useCompression=true)
   url: null
+  # -- External activemq username
   user: null
+  # -- External activemq password
   password: null
```
alxgomz, line 356: I don't think we should mention "external" here. All broker are external as managed by the umbrella chart

alxgomz (changes requested):
I'm wondering if it's really usefull to offer extra valumoes and volume mounts for t-engines and sfs. Do you have a particular use-case in mind?

.github/workflows/lint-test.yaml
```diff
@@ -92,6 +92,13 @@ jobs:
...
+          --namespace "$INSTALL_NAMESPACE"
+          $TEST_ALL_CHARTS_ARG
+          --skip-clean-up
+
+      - name: Debug on failure
+        if: failure()
         run: |
-          ct install --config ct-install.yaml \
-            --namespace "$INSTALL_NAMESPACE" $TEST_ALL_CHARTS_ARG
+          kubectl --namespace "$INSTALL_NAMESPACE" get svc
```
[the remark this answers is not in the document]
gionn: to debug connection issues between components, ct is only describing pods which didn't help in this case

charts/alfresco-transform-service/values.yaml
gionn: I am removing completely the ai chart dependency then

charts/alfresco-transform-service/templates/_helpers-name.tpl
gionn: indeed I can reuse the same named template for deployment/config/secret and have a separated name template only for additional config/secrets

gionn:
> I'm wondering if it's really usefull to offer extra valumoes and volume mounts for t-engines and sfs. Do you have a particular use-case in mind?

java keystores

charts/alfresco-transform-service/templates/_helpers-name.tpl
```diff
@@ -1,69 +1,44 @@
-{{- define "alfresco-transform-service.config-filestore.name" -}}
-{{- $scope := (dict "Values" (dict "nameOverride" "filestore-configmap" ) "Chart" .Chart "Release" .Release) }}
+{{- define "alfresco-transform-service.messagebroker.name" -}}
+{{- $scope := (dict "Values" (dict "nameOverride" "messagebroker" ) "Chart" .Chart "Release" .Release) }}
```
gionn, line 2: did I risk a conflict when used under the umbrella chart?

alxgomz: strictly speaking no as this name is not used elsewhere for now but there is no reason why such a cm would be used by this chart rather than another which also leverages the activemq component.

PR #177, opened 2024-01-03
OPSEXP-2293: add AI pipelines as configmap

alxgomz:
Ref: OPSEXP-2293

### DIFF
--- a/charts/alfresco-ai-transformer/templates/deployment-ai-transformer.yaml
+++ b/charts/alfresco-ai-transformer/templates/deployment-ai-transformer.yaml
@@ -20,40 +20,7 @@ spec:
     spec:
       serviceAccountName: {{ include "alfresco-ai-transformer.serviceAccountName" . }}
       {{- include "alfresco-common.component-pod-security-context" .Values | indent 4 }}
-      {{- with .Values.nodeSelector }}
-      nodeSelector:
-        {{- .Values.nodeSelector | toYaml | nindent 8 }}
-      {{- end }}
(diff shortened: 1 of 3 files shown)


charts/alfresco-ai-transformer/templates/deployment-ai-transformer.yaml
```diff
@@ -20,8 +20,9 @@ spec:
     spec:
       serviceAccountName: {{ include "alfresco-ai-transformer.serviceAccountName" . }}
       {{- include "alfresco-common.component-pod-security-context" .Values | indent 4 }}
-      {{- if .Values.nodeSelector }}
-      nodeSelector: {{- .Values.nodeSelector | toYaml | nindent 8 }}
+      {{- with .Values.nodeSelector }}
+      nodeSelector:
+        {{- .Values.nodeSelector | toYaml | nindent 8 }}
```
gionn, line 25: move it to the bottom of the template together with toleration and affinity, like in https://github.com/Alfresco/alfresco-helm-charts/blob/main/charts/alfresco-search-enterprise/templates/liveindexing-deployment.yaml#L78 (I was following the same ordering of the helm boilerplate when you generate a new chart, should help when we make a new named template)

gionn (approved):

charts/alfresco-ai-transformer/templates/deployment-ai-transformer.yaml
alxgomz: I had to move the `affinity` value to a string to cope with default & comon uses cases

PR #179, opened 2024-01-04
OPSEXP-2291 Bump Search services chart to most recent conventions

gionn:
Ref: OPSEXP-2291


charts/alfresco-search-service/templates/secret-repository.yaml
```diff
@@ -1,9 +1,8 @@
-{{- if (not .Values.repository.existingSecret.name) }}
-{{- if and (eq "secret" .Values.repository.securecomms) (not .Values.repository.existingConfigMap.name) }}
+{{- if and (not .Values.repository.existingSecret.name) (eq "secret" .Values.repository.securecomms) }}
```
gionn, line 1: double check, I removed `not .Values.repository.existingConfigMap.name`

charts/alfresco-search-service/charts/alfresco-insight-zeppelin/templates/deployment.yaml
```diff
@@ -3,44 +3,56 @@
...
-      app: {{ template "alfresco-insight-zeppelin.fullName" . }}
-      release: {{ .Release.Name }}
+      {{- include "alfresco-insight-zeppelin.selectorLabels" . | nindent 6 }}
   strategy:
     type: RollingUpdate
     rollingUpdate: {{- toYaml .Values.global.strategy.rollingUpdate | nindent 6 }}
   template:
     metadata:
       annotations:
         checksum/config: {{ include (print $.Template.BasePath "/config.yaml") . | sha256sum }}
```
alxgomz, line 20: please remove

charts/alfresco-search-service/charts/alfresco-insight-zeppelin/templates/config.yaml
```diff
@@ -1,17 +1,15 @@
...
-    component: zeppelin
+    {{- include "alfresco-insight-zeppelin.labels" . | nindent 4 }}
 data:
-  REPO_HOST: "{{ template "alfresco-repo-host" . }}"
-  REPO_PORT: "{{ template "alfresco-repo-port" . }}"
-  {{- range $key, $value := .Values.environment }}
-  {{ $key }}: {{ $value }}
+  {{- with .Values.repository }}
+  REPO_HOST: >-
+    {{ template "alfresco-common.url.host" .url }}
```
alxgomz, line 11: So we have different config keys for values based configuration or configmap based config. I know this is what was initially wriiten in thez JIRA but I'm wondering if it's a good idea

charts/alfresco-search-service/templates/secret-repository.yaml
alxgomz: I think it was there because if the repo is configured through a cm then the securecomms is not accessible to check whether is set to `secret`.  So if the cm is configured with `securecomms=none` the secret would still be created by the chart (as the default value set i s `seret`)

alxgomz (changes requested):

charts/alfresco-search-service/templates/secret-repository.yaml
gionn: at this point I am not sure it's a good idea to have `securecomms` inside the configmap, what's the reason for it? it's not something dynamic like hostname/port

charts/alfresco-search-service/charts/alfresco-insight-zeppelin/templates/deployment.yaml
gionn: why? we have a configmap and should help if changing it

alxgomz (changes requested):
I will insist on getting rid of the config/checksum which I really don't like

alxgomz (approved):

PR #181, opened 2024-01-08
OPSEXP-2300: align alfresco-sync-service charts to latest conventions

alxgomz:
Ref: OPSEXP-2300

* adds support for message broker existing configmap
* adds support for message broker existing secret
* adds support for db existing configmap
* adds support for db existing secret
* adds support for repo existing configmap
* fix named templates
* remove recursive usage of  `nameOverride`
* add service account
* fix resource naming


charts/alfresco-sync-service/templates/deployment.yaml
```diff
@@ -0,0 +1,139 @@
...
+        {{- end }}
+      annotations:
+        {{- with .Values.podAnnotations }}
+          {{- toYaml . | nindent 8 }}
+        {{- end }}
+    spec:
+      {{- include "alfresco-common.component-pod-security-context" .Values | indent 4 }}
+      serviceAccountName: {{ include "alfresco-sync-service.serviceAccountName" . }}
+      {{- include "alfresco-common.imagePullSecrets" . | indent 6 }}
+      terminationGracePeriodSeconds: 60
```
gionn, line 32: docker image is not handling correctly shutdown requests?

worth having a value to configure it imho

charts/alfresco-sync-service/templates/deployment.yaml
```diff
@@ -0,0 +1,139 @@
...
+      - name: syncservice
+        image: "{{ .Values.image.repository }}:{{ .Values.image.tag }}"
+        imagePullPolicy: {{ .Values.image.pullPolicy }}
+        {{- include "alfresco-common.component-security-context" .Values | indent 6 }}
+        envFrom:
+        - configMapRef:
+            name: {{ template "alfresco-sync-service.fullname" . }}
+        env:
+          {{- with .Values.database }}
+          {{- $dbCtx := dict "Values" (dict "nameOverride" "alfresco-sync-database") "Chart" $.Chart "Release" $.Release }}
```
gionn, line 43: I don't like having the duplication of the nameOverride value, better have a dedicated named template (e.g. `.name` in _helpers-name.tpl)

charts/alfresco-sync-service/templates/ingress.yaml
```diff
@@ -3,7 +3,7 @@
 apiVersion: {{ template "common.capabilities.ingress.apiVersion" . }}
 kind: Ingress
 metadata:
-  name: {{ template "alfresco-sync-service.fullname" . }}-ingress
+  name: {{ template "alfresco-sync-service.fullname" . }}
   annotations:
     kubernetes.io/ingress.class: "nginx"
```
gionn, line 8: check the helm install logs, I am switching to `ingressClassName: nginx` in search service PR

charts/alfresco-sync-service/templates/ingress.yaml
```diff
@@ -3,7 +3,7 @@
 apiVersion: {{ template "common.capabilities.ingress.apiVersion" . }}
```
gionn, line 3: I guess we can get rid of this

charts/alfresco-sync-service/templates/deployment.yaml
alxgomz: I'm not sure tbh. I've just kept what was there before. But after taking a quick look it seems the docker image is using a wrapper script that does some substitutions before spawning the server.... So I can only imagine that this is indeed not handled properly by docker

charts/alfresco-sync-service/values.yaml
```diff
@@ -54,83 +54,79 @@ readinessProbe:
...
+  password: null
+  existingConfigMap:
+    name: null
+    keys:
+      # -- configmap key where to find the URL of the database
+      url: DATABASE_URL
+      # -- configmap key where to find the JDBC driver class to use.
+      # The configmap may leverage the alfresco-repository.db.cm named
+      # template to auto-generate it from the sole url parameter.
+      driver: DATABASE_DRIVER
```
gionn, line 112: not sure it make sense to have driver here but I understand is correlated to url so if it change then you may have to provide a different driver

gionn (approved):

PR #189, opened 2024-01-15
OPSEXP-2440: use static cm name or easier ATS trouter config

alxgomz:
Ref: OPSEXP-2440

### DIFF
--- a/charts/alfresco-ai-transformer/templates/config-ai-transform-pipelines.yaml
+++ b/charts/alfresco-ai-transformer/templates/config-ai-transform-pipelines.yaml
@@ -1,8 +1,7 @@
 apiVersion: v1
 kind: ConfigMap
 metadata:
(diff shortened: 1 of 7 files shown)


charts/alfresco-ai-transformer/templates/config-ai-transform-pipelines.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: v1
 kind: ConfigMap
 metadata:
-  {{- $cmCtx := dict "Values" (dict "nameOverride" "ai-pipelines") "Chart" .Chart "Release" .Release }}
+  {{- $cmCtx := dict "Values" (dict "fullnameOverride" "ai-transform-pipelines") "Chart" .Chart "Release" .Release }}
```
gionn, line 4: it still make sense to use the named template with fullnameOverride? just hardcoded it then (or define a new value)

alxgomz: not much indeed

PR #193, opened 2024-01-16
OPSEXP-2444 Bump versions for Draco release

gionn:
requires https://github.com/Alfresco/alfresco-updatecli/pull/24

Ref: OPSEXP-2444

### DIFF
--- a/charts/alfresco-search-service/values.yaml
+++ b/charts/alfresco-search-service/values.yaml
@@ -17,14 +17,14 @@ podSecurityContext:
   fsGroup: 33007
 searchServicesImage:
   repository: quay.io/alfresco/search-services
-  tag: 2.0.8.2
+  tag: 2.0.9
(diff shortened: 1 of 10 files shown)


charts/alfresco-search-service/values.yaml
```diff
@@ -132,15 +127,12 @@ initContainer:
...
-
 global:
   alfrescoRegistryPullSecrets: quay-registry-secret
-
 nodeSelector: {}
 tolerations: []
 # -- Pod affinity, passed thru tpl function
-affinity: |-
-
+affinity: ""
```
gionn, line 135: raised https://github.com/updatecli/updatecli/issues/1875

alxgomz: I'm fine with that

alxgomz (approved):

PR #197, opened 2024-01-17
OPSEXP-2444 Cleanup search services updatecli pipeline

gionn:
* Drop unnecessary custom pipeline for search services (already covered by the alfresco-updatecli pipeline)
* Ignore external/test dependencies
* Ignore insight-zeppelin to workaround https://github.com/updatecli/updatecli/issues/1876

Ref: OPSEXP-2444


alxgomz (commented):
But then how do we cover the search updates? this chart is explicitly excluded from the default pipeline which do not support the custom images path IIRC

gionn:
> But then how do we cover the search updates? this chart is explicitly excluded from the default pipeline which do not support the custom images path IIRC

default pipeline is just for help dependencies, tags are handled via https://github.com/Alfresco/alfresco-updatecli/blob/master/deployments/uber-manifest.tpl

alxgomz (changes requested):
I forgot we introduced the `ignorecontainer` params. In that case we can we also remove the `autodiscovery.crawlers.helm.ignore` entries running updatecli in diff mode seems to not shock on search folder

alxgomz (approved):

PR #198, opened 2024-01-17
OPSEXP-2444: Bump search enterprise chart to 3.1.0

gionn:
Ref: OPSEXP-2444


alxgomz (approved):
Can you change the PR title so its show ASE chart has been bumped?

PR #207, opened 2024-01-31
OPSEXP-2452 Support share clustering

gionn:
Ref: OPSEXP-2452

### DIFF
--- a/charts/alfresco-share/templates/configmap-custom-application-context.yaml
+++ b/charts/alfresco-share/templates/configmap-custom-application-context.yaml
@@ -1,4 +1,4 @@
-{{- if gt .Values.replicaCount 1 -}}
+{{- if gt (int .Values.replicaCount) 1 -}}
 apiVersion: v1
 kind: ConfigMap
 metadata:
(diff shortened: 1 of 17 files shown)


charts/alfresco-share/templates/configmap-custom-application-context.yaml
```diff
@@ -0,0 +1,36 @@
...
+      <hz:topic id="topic" instance-ref="webframework.cluster.slingshot" name="slingshot-topic"/>
+      <hz:hazelcast id="webframework.cluster.slingshot">
+          <hz:config>
+            <hz:network port="{{ .Values.hazelcast.port }}" port-auto-increment="false">
+                <hz:join>
+                  <hz:kubernetes enabled="true" namespace="{{ .Release.Namespace }}" service-name="{{ template "alfresco-share.service-hz.name" . }}" />
+                </hz:join>
+                <hz:interfaces enabled="false">
+                  <hz:interface>110.42.0.*</hz:interface>
+                </hz:interfaces>
```
alxgomz, line 27: No need to keep that as it's disabled

alxgomz (changes requested):
Looks pretty good, though this would only work starting  from 23.1, we'd need to either:

* make it conditional to `appVersion` (not sure how robust it is as people might customize it)
* make it more generic generic extra volumes and so on and providing documentation (ofc far less easy to use)

charts/alfresco-share/templates/configmap-custom-application-context.yaml
gionn: sorry, just forgot to drop after testing it

alxgomz (approved):

PR #211, opened 2024-02-02
OPSEXP-2488 Configurable ingress class in repository chart

pmacius:
- Drop support for previous apiVersions

Ref: [OPSEXP-2488](https://alfresco.atlassian.net/browse/OPSEXP-2488)

[OPSEXP-2488]: https://alfresco.atlassian.net/browse/OPSEXP-2488?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/charts/alfresco-repository/tests/ingress_test.yaml
+++ b/charts/alfresco-repository/tests/ingress_test.yaml
@@ -21,25 +21,6 @@ tests:
   - it: should sanitize ingress
     capabilities:
       majorVersion: 1
-      minorVersion: 17
-    set:
-      ingress:
-        annotations:
-          kubernetes.io/ingress.class: myfancyClass
-          nginx.ingress.kubernetes.io/server-snippet: listen 6666;
-    asserts:
-      - notMatchRegex:
-          path: metadata.annotations['nginx.ingress.kubernetes.io/server-snippet']
-          pattern: listen 6666;
-        template: ingress.yaml
-      - equal:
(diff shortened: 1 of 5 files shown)


charts/alfresco-repository/tests/ingress_test.yaml
```diff
@@ -19,28 +19,6 @@ tests:
         template: ingress.yaml
```
gionn, line 19: you can remove all the template inside each test since this suite include only one template

charts/alfresco-repository/tests/ingress_test.yaml
gionn, line 1: you can add a test which cover also the contents of tls and rules which I don't see them tested

charts/alfresco-repository/values.yaml
```diff
@@ -214,6 +214,7 @@ ingress:
   enabled: true
   # -- provide annotations for nginx ingress (no toher ingress is supported at
   # that point)
+  className: nginx
```
gionn, line 217: documentation is automatically generated by the precededing `# --` so make sure so move it either after annotations or before the comment

you can add a comment to className as well

gionn (changes requested):

gionn (approved):

alxgomz (approved):
👏

PR #213, opened 2024-02-06
OPSEXP-2443 Add recommended labels to all charts

pmacius:
OPSEXP-2443

### DIFF
--- a/charts/activemq/Chart.yaml
+++ b/charts/activemq/Chart.yaml
@@ -7,7 +7,7 @@ keywords:
 name: activemq
 sources:
   - https://github.com/Alfresco/alfresco-helm-charts
-version: 3.4.2
+version: 3.5.0
 appVersion: 5.18.3
 dependencies:
   - name: alfresco-common
--- a/charts/activemq/Chart.yaml
+++ b/charts/activemq/Chart.yaml
@@ -7,7 +7,7 @@ keywords:
 name: activemq
 sources:
   - https://github.com/Alfresco/alfresco-helm-charts
-version: 3.4.1
+version: 3.4.2
 appVersion: 5.18.3
 dependencies:
   - name: alfresco-common
--- a/charts/activemq/templates/_helpers.tpl
+++ b/charts/activemq/templates/_helpers.tpl
@@ -40,6 +40,7 @@ helm.sh/chart: {{ include "activemq.chart" . }}
 app.kubernetes.io/version: {{ .Chart.AppVersion | quote }}
 {{- end }}
 app.kubernetes.io/managed-by: {{ .Release.Service }}
+app.kubernetes.io/component: {{ template "activemq.name" . }}
 {{- end }}
 
 {{/*
@@ -48,7 +49,6 @@ Selector labels
 {{- define "activemq.selectorLabels" -}}
 app.kubernetes.io/name: {{ include "activemq.name" . }}
 app.kubernetes.io/instance: {{ .Release.Name }}
-app.kubernetes.io/component: {{ template "activemq.name" . }}
 {{- end }}
 
 {{/*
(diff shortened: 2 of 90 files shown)


charts/activemq/templates/_helpers.tpl
```diff
@@ -48,6 +48,7 @@ Selector labels
 {{- define "activemq.selectorLabels" -}}
 app.kubernetes.io/name: {{ include "activemq.name" . }}
 app.kubernetes.io/instance: {{ .Release.Name }}
+app.kubernetes.io/component: {{ template "activemq.name" . }}
```
gionn, line 51: technically it will not change the fact that component will have the proper component label set, but conceptually will use it as a selector in the service and that's not the purpose of this additional label

it's better if you add it in the `.labels` named template

gionn, line 51: I think we should avoid using any `.name` named template, if you check the current implementation you see that is namespaced with the helm release name, while I imagine we want a fixed string which identity the component name (multiple helm install should yield the same component name).

I guess using `{{ .Chart.Name }}` should work

charts/activemq/Chart.yaml
```diff
@@ -7,7 +7,7 @@ keywords:
 name: activemq
 sources:
   - https://github.com/Alfresco/alfresco-helm-charts
-version: 3.4.1
+version: 3.4.2
```
gionn, line 10: we probably want to bump the minor version (the second digit) given that we are supporting a new thing that was not supported before, and not really a bug fix (something that was supposed to be working but it was not)

gionn (changes requested):

charts/alfresco-search-enterprise/templates/configmap-ats.yaml
```diff
@@ -6,6 +6,7 @@ metadata:
   name: {{ template "alfresco-search-enterprise.fullname" $ctx }}
   labels:
     {{- include "alfresco-search-enterprise.labels" . | nindent 4 }}
+    app.kubernetes.io/component: {{ include "alfresco-search-enterprise.name" . }}
```
gionn, line 9: why you didn't extend `alfresco-search-enterprise.labels` here?

charts/alfresco-search-enterprise/templates/reindexing-job.yaml
```diff
@@ -14,6 +15,7 @@ spec:
       {{- end }}
       labels:
         {{- include "alfresco-search-enterprise.selectorLabels" . | nindent 8 }}
+        app.kubernetes.io/component: {{ include "alfresco-search-enterprise.name" . }}
```
gionn, line 18: this usage of selectorLabels instead of labels really looks like a bug

use https://github.com/gionn/helm-new-chart/blob/master/templates/deployment.yaml as a reference - it's the boiler template when asking helm to create a new chart

charts/alfresco-search-enterprise/templates/_helpers.tpl
```diff
@@ -40,6 +40,7 @@ helm.sh/chart: {{ include "alfresco-search-enterprise.chart" . }}
 app.kubernetes.io/version: {{ .Chart.AppVersion | quote }}
 {{- end }}
 app.kubernetes.io/managed-by: {{ .Release.Service }}
+app.kubernetes.io/component: {{ include "alfresco-search-enterprise.name" . }}
```
gionn, line 43: at this point we probably want to avoid using this named template here and just use `.Chart.Name`

charts/alfresco-transform-service/templates/_helpers.tpl
```diff
@@ -40,6 +40,7 @@ helm.sh/chart: {{ include "alfresco-transform-service.chart" . }}
 app.kubernetes.io/version: {{ .Chart.AppVersion | quote }}
 {{- end }}
 app.kubernetes.io/managed-by: {{ .Release.Service }}
+app.kubernetes.io/component: {{ template "alfresco-transform-service.name" . }}
```
gionn, line 43: add some tests for this chart because I imagine it has the same problem (using the overridden name by the custom labals named template) I've just spotted before

gionn (changes requested):

gionn (approved):

PR #215, opened 2024-02-12
OPSEXP-2437 Avoid path key in transformers configmaps

gionn:
Ref: OPSEXP-2437

### DIFF
--- a/charts/alfresco-transform-service/templates/config-libreoffice.yaml
+++ b/charts/alfresco-transform-service/templates/config-libreoffice.yaml
@@ -8,7 +8,7 @@ metadata:
 data:
   {{- include "alfresco-transform-service.sfs.url" . | indent 2 }}
   {{- toYaml .Values.libreoffice.environment | nindent 2 }}
-  {{- range $key, $val := .Values.libreoffice.livenessProbe }}
+  {{- range $key, $val := omit .Values.libreoffice.livenessProbe "path" }}
   {{ $key }}: {{ $val | quote }}
   {{- end }}
 {{- end }}
--- a/charts/alfresco-transform-service/templates/config-transform-router.yaml
+++ b/charts/alfresco-transform-service/templates/config-transform-router.yaml
@@ -16,7 +16,4 @@ data:
   MISC_URL: http://{{ template "alfresco-transform-service.transform-misc.name" . }}:80
   {{- end }}
   {{- toYaml .Values.transformrouter.environment | nindent 2 }}
-  {{- range $key, $val := .Values.transformrouter.livenessProbe }}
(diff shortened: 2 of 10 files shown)


charts/alfresco-transform-service/templates/config-transform-router.yaml
```diff
@@ -16,7 +16,4 @@ data:
   MISC_URL: http://{{ template "alfresco-transform-service.transform-misc.name" . }}:80
   {{- end }}
   {{- toYaml .Values.transformrouter.environment | nindent 2 }}
-  {{- range $key, $val := .Values.transformrouter.livenessProbe }}
-  {{ $key }}: {{ $val | quote }}
-  {{- end }}
```
alxgomz, line 21: I guess this is because router do not have that logic of transformation performance for its probes but in that we should also remove linked values

charts/alfresco-transform-service/templates/config-libreoffice.yaml
```diff
@@ -8,7 +8,7 @@ metadata:
 data:
   {{- include "alfresco-transform-service.sfs.url" . | indent 2 }}
   {{- toYaml .Values.libreoffice.environment | nindent 2 }}
-  {{- range $key, $val := .Values.libreoffice.livenessProbe }}
+  {{- range $key, $val := omit .Values.libreoffice.livenessProbe "path" }}
```
alxgomz, line 11: Why did we remove it from router but not from sfs?

alxgomz (changes requested):

charts/alfresco-transform-service/templates/config-transform-router.yaml
gionn: values are already good:
```
  readinessProbe:
    path: /actuator/health
    initialDelaySeconds: 20
    periodSeconds: 60
    timeoutSeconds: 10
  livenessProbe:
    path: /actuator/health
    initialDelaySeconds: 20
    periodSeconds: 30
    timeoutSeconds: 10
```

charts/alfresco-transform-service/templates/config-libreoffice.yaml
gionn: good catch, but please write comments on the appropriate line :D

alxgomz: oops :p

charts/alfresco-transform-service/templates/config-transform-router.yaml
alxgomz: OK I missed the were expanded in the deplopyment

alxgomz (approved):

PR #235, opened 2024-02-27
OPSEXP-2502 Add required security context to adf app chart

gionn:
`runAsUser: 101` is required for adw images shipped with 7.3 and 7.2 as per failures in:

https://github.com/Alfresco/acs-deployment/actions/runs/8052691173/job/21993419141

Ref: OPSEXP-2502


charts/alfresco-adf-app/values.yaml
```diff
@@ -23,9 +23,16 @@ serviceAccount:
...
+podSecurityContext:
+  runAsUser: 101
+  runAsNonRoot: true
+
+securityContext:
+  allowPrivilegeEscalation: false
+  capabilities:
+    drop:
+      - NET_RAW
+      - ALL
```
pmacius, line 35: Is `NET_RAW` not a part of `ALL`?

gionn: good catch, I simply copy-pasted what was in our common library, will create a task

charts/alfresco-adf-app/values.yaml
```diff
@@ -23,9 +23,16 @@ serviceAccount:
...
-securityContext: {}
+podSecurityContext:
+  runAsUser: 101
+  runAsNonRoot: true
+
+securityContext:
+  allowPrivilegeEscalation: false
+  capabilities:
+    drop:
+      - NET_RAW
```
gionn, line 34:
```suggestion
```

alxgomz (approved):

PR #238, opened 2024-03-04
OPSEXP-2519 Bump version for 23.2 release

pmacius:
!Ref OPSEXP-2519


alxgomz (changes requested):
appVersion bump should trigger minor char version bump instead of bugfix.

gionn (approved):

alxgomz (approved):

PR #243, opened 2024-03-06
OPSEXP-2514 Add helm hooks on reindexing job to make it rerun on installation.

pmacius:
!Ref OPSEXP-2514

### DIFF
--- a/charts/alfresco-search-enterprise/values.yaml
+++ b/charts/alfresco-search-enterprise/values.yaml
@@ -65,6 +65,12 @@ reindexing:
         # -- Key within the configmap holding the full url to connect to the
         # alfresco repository
         url: REPOSITORY_URL
+  # -- Helm hooks definition
+  policy:
+    # -- When the job will be executed
+    run: post-install
+    # -- When the resources associated with job will be deleted. By default resources will be kept until a new installation.
+    delete: before-hook-creation
   db:
     # -- Provide the full JDBC url to connect to database service e.g.:
     # `jdbc:postgresql://hostname:5432/database`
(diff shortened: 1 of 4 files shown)


charts/alfresco-search-enterprise/values.yaml
```diff
@@ -65,6 +65,12 @@ reindexing:
         # -- Key within the configmap holding the full url to connect to the
         # alfresco repository
         url: REPOSITORY_URL
+  # -- Helm hooks definition
+  policy:
+    # -- When the job will be executed
+    run: post-install
+    # -- When the resources associated with job will be deleted. By default resources will be kept until a new installation.
+    delete: before-hook-creation
```
gionn, line 73: this is the default as per [docs](https://helm.sh/docs/topics/charts_hooks/#hook-deletion-policies), so unless we find it's meaningful to fine tune it, there is no need to have a dedicated value for it

charts/alfresco-search-enterprise/values.yaml
```diff
@@ -65,6 +65,12 @@ reindexing:
         # -- Key within the configmap holding the full url to connect to the
         # alfresco repository
         url: REPOSITORY_URL
+  # -- Helm hooks definition
```
gionn, line 68:
```suggestion
```

usually we document only leaf nodes to avoid having the additional entry in the docs with the entire payload as example

charts/alfresco-search-enterprise/values.yaml
```diff
@@ -65,6 +65,12 @@ reindexing:
...
+  policy:
+    # -- When the job will be executed
+    run: post-install
```
gionn, line 71: `policy` doesn't look quite descriptive of what we are doing here.

`hook` as top level key and `name` for the hook event?

PR #249, opened 2024-03-14
OPSEXP-2488 Fix apiVersion of ingresses for msteams and ms365

pmacius:
Ref: OPSEXP-2488


gionn (commented):
my eye slipped down to the annotations section and I am afraid it won't work if more than one annotation is set, can you check/write a quick unit test?

it should be:
```
{{- toYaml .Values.ingress.extraAnnotations | nident 4 }}
```

gionn (approved):

PR #257, opened 2024-03-26
OPSEXP-2592: alfresco-process-service - update ingress template to match other charts

alxgomz:
Ref: OPSEXP-2592

### DIFF
--- a/charts/alfresco-process-services/values.yaml
+++ b/charts/alfresco-process-services/values.yaml
@@ -5,11 +5,6 @@
 global:
   alfrescoRegistryPullSecrets: quay-registry-secret
 
-ingress:
-  enabled: true
-  protocol: http
-  hostName: ''
-
 processEngine:
   replicaCount: 1
   podLabels: {}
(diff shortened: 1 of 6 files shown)


charts/alfresco-process-services/values.yaml
```diff
@@ -146,9 +156,23 @@ adminApp:
...
-    maxUploadSize: "5G"
+    # -- Toggle ingress for APS application
+    enabled: true
+    # -- provide annotations for nginx ingress (no toher ingress is supported at
+    # that point)
+    annotations:
+      nginx.ingress.kubernetes.io/proxy-body-size: 5g
+      nginx.ingress.kubernetes.io/affinity: "cookie"
+    # -- supported ingress class
+    className: nginx
```
gionn, line 167: check typo and not sure why we need to explicit what we support here, it's pretty self-explanatory and doesn't prevent users to provide their own config

gionn (changes requested):

charts/alfresco-process-services/values.yaml
alxgomz: right that's actually a copy/paste from repo

gionn (approved):

pmacius (approved):

PR #260, opened 2024-03-29
OPSEXP-2601: split APS charts into 2 activit component charts

alxgomz:
Ref: OPSEXP-2601


charts/alfresco-activiti/Chart.yaml
```diff
@@ -0,0 +1,15 @@
+apiVersion: v2
+description: A Helm chart for Alfresco Activiti
+name: alfresco-activiti
```
gionn, line 3: are those the best names?

I see that on docs https://docs.alfresco.com/process-services/latest/install/manual/

we are calling them process services and process services administrator

alxgomz: names validated with PM Arabella. @vprovaggi was in the chat but did not reply yet

gionn (approved):

pmacius (approved):

PR #261, opened 2024-03-29
OPSEXP-2525 alfresco-transform-service introduce optional hpa

pmacius:
Ref: OPSEXP-2525


pmacius:
Please add your suggestions. Since pods for ats are ready pretty quick I changed the readiness probe and added smaller stabilization window for scaling down. Not sure if it is good to keep comments in values.

charts/alfresco-transform-service/values.yaml
```diff
@@ -130,6 +130,43 @@ pdfrenderer:
     maxTransformSeconds: 1200
   volumes: []
   volumeMounts: []
+  autoscaling:
+    # -- Toggle pdfrenderer autoscaling
+    enabled: false
+    # -- minimum number of replicas to spin up within the replicatset
+    minReplicas: 1
+    # -- maximum number of replicas to spin up within the replicatset
```
gionn, line 138:
```suggestion
    # -- minimum number of replicas to spin up within the replicaset
    minReplicas: 1
    # -- maximum number of replicas to spin up within the replicaset
```

fix also in the others sections :D

alxgomz (commented):
what's the reason for using 1m as a scaledown stabilization window? I find it short but this is based on thin air tbh.

pmacius:
> what's the reason for using 1m as a scaledown stabilization window? I find it short but this is based on thin air tbh.
@alxgomz 
Mainly because I think that for those deployments default scale down window is to long (5 minutes). It can scale up pretty fast so user won't experience much difference and we can save more resources. Also looking at the prometheus graphs on develop cluster its a straight line so from our perspective it is hard to guess which would be the best period of time. From here I think we can also keep it default without changing and give some comments about changing it depending on how much this resource is used.

alxgomz (approved):

PR #264, opened 2024-04-03
OPSEXP-2601: cleanup chart index

alxgomz:
Ref: OPSEXP-2601


pmacius (approved):

gionn (commented):
why?

alxgomz:
> why?

For the sake of having clean index which do not contain ref to charts that would exist anymore

gionn (approved):

PR #266, opened 2024-04-08
OPSEXP-2529: ACTIVITI-5113 workaround

alxgomz:
Ref: OPSEXP-2529

### DIFF
--- a/charts/alfresco-activiti-admin/values.yaml
+++ b/charts/alfresco-activiti-admin/values.yaml
@@ -82,6 +82,12 @@ database:
       # -- Key within the secret holding the database password for admin app
       password: DATABASE_PASSWORD
 
+workaround:
+  # -- Enable workaround for https://hyland.atlassian.net/browse/ACTIVITI-5113
+  # Set to `true` to enable the workaround regardless of the appVersion
+  # Set to `false` to disable the workaround (or any other value)
+  ACTIVITI-5113: true
+
 tags:
   # -- A chart tag used for Hyland's CI purpose. Do not set it to true.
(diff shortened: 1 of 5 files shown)


charts/alfresco-activiti-admin/values.yaml
```diff
@@ -82,6 +82,12 @@ database:
       # -- Key within the secret holding the database password for admin app
       password: DATABASE_PASSWORD
 
+workaround:
+  # -- Enable workaround for https://hyland.atlassian.net/browse/ACTIVITI-5113
+  # Set to `true` to enable the workaround regardless of the appVersion
```
gionn, line 87:
```suggestion
  # Set to `true` to enable the workaround
```

why mention appVersion if there is not logic around it?

alxgomz: Leftovers! Thank you/

gionn (approved):

charts/alfresco-activiti-admin/values.yaml
```diff
@@ -82,6 +82,12 @@ database:
...
+  # Set to `false` to disable the workaround (or any other value)
+  ACTIVITI-5113: true
```
pmacius, line 89: I am not sure if this is a good value name. Maybe `activiti-entrypoint` would be more descriptive. It would give some idea on what type of workaround is this.

alxgomz: I think it's difficult to describe the workaround itself so linking to the jira seemed like a good enough choice.

PR #288, opened 2024-05-06
OPSEXP-2620 Generate documentation website with jekyll

gionn:
Ref: OPSEXP-2620

already live at: https://alfresco.github.io/alfresco-helm-charts/

requires:
* https://github.com/Alfresco/alfresco-helm-charts/pull/292
* https://github.com/Alfresco/alfresco-helm-charts/pull/293
* https://github.com/Alfresco/jekyll-build-tools/pull/3


_config.yml
```diff
@@ -0,0 +1,48 @@
...
+url: https://alfresco.github.io/alfresco-helm-charts
+
+aux_links:
+  Docs: https://docs.alfresco.com/
+  GitHub Repository: https://github.com/Alfresco/alfresco-helm-charts
+
+aux_links_new_tab: true
+
+gh_edit_link: true
+gh_edit_link_text: "Improve this page on GitHub"
```
pmacius, line 15: This link redirects to https://github.com/Alfresco/alfresco-helm-charts/edit/main/docs/index.md which doesn't exist.

pmacius:
Found just this one issue, besides that looking great. If it's not a quick fix I think we can create Bug for this.

_config.yml
gionn: yet, it's just because I renamed that file and is not yet on main

alxgomz (approved):

.github/workflows/docs.yml
```diff
@@ -0,0 +1,29 @@
...
+      - _config.yml
+      - Gemfile*
+      - charts/**/README.md
+  workflow_run:
+    workflows: [Release Charts]
+    types: [completed]
+
+jobs:
+  publish:
+    uses: Alfresco/jekyll-build-tools/.github/workflows/jekyll-publish.yml@OPSEXP-2620-gh-pages
```
gionn, line 24:
```suggestion
    uses: Alfresco/jekyll-build-tools/.github/workflows/jekyll-publish.yml@4a682d36e45971f30e235c28f7b8a0898574b1b7
```

PR #314, opened 2024-06-12
OPSEXP-2648 Integrate aps into updatecli matrix

pmacius:
Ref: OPSEXP-2648


alxgomz (approved):
Should be OK once updatecli matrix PR is merged

PR #320, opened 2024-06-18
OPSEXP-2664: update APS app probes endpoints

alxgomz:
Ref: OPSEXP-2664

(APS admin dev images are not released so we'll have to wait for release before we can do/test it OPSEXP-2702)

### DIFF
--- a/charts/alfresco-activiti/README.md
+++ b/charts/alfresco-activiti/README.md
@@ -64,11 +64,12 @@ A Helm chart for Alfresco Activiti
 | ingress.hosts[0].paths[0].pathType | string | `"Prefix"` |  |
 | ingress.tls | list | `[]` |  |
 | license.secretName | string | `nil` |  |
-| livenessProbe.failureThreshold | int | `5` |  |
-| livenessProbe.initialDelaySeconds | int | `25` |  |
-| livenessProbe.path | string | `"/activiti-app/app/rest/locale"` |  |
-| livenessProbe.periodSeconds | int | `10` |  |
-| livenessProbe.timeoutSeconds | int | `5` |  |
+| livenessProbe.<<.failureThreshold | int | `5` |  |
+| livenessProbe.<<.initialDelaySeconds | int | `25` |  |
+| livenessProbe.<<.path | string | `"/activiti-app/actuator/health/readiness"` | Make sure to set `readinessProbe.path` to `/activiti-app/app/rest/locale` if you're using a version of APS prior to 24.2.0 |
+| livenessProbe.<<.periodSeconds | int | `10` |  |
+| livenessProbe.<<.timeoutSeconds | int | `5` |  |
+| livenessProbe.path | string | `"/activiti-app/actuator/health/liveness"` |  |
 | nodeSelector | object | `{}` |  |
 | persistence.accessModes[0] | string | `"ReadWriteOnce"` |  |
 | persistence.baseSize | string | `"20Gi"` | Initial default size of dynamically provisioned storage |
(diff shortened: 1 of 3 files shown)


charts/alfresco-activiti/values.yaml
```diff
@@ -61,12 +61,16 @@ persistence:
...
+  path: /activiti-app/actuator/health/readiness
   initialDelaySeconds: 25
   periodSeconds: 10
   failureThreshold: 5
   timeoutSeconds: 5
-livenessProbe: *apsProbe
+livenessProbe:
+  <<: *apsProbe
+  # -- Make sure to set `readinessProbe.path` to `/activiti-app/app/rest/locale` if you're using a version of APS prior to 24.3.0
+  path: /activiti-app/actuator/health/liveness
```
pmacius, line 73: Shouldn't we have some different initialDelayseconds etc. now when the endpoints are different

alxgomz: I see no good settings I could change here. even the initial delay is not worth changing given the app runs in tomcat and tomcat would only treat requests after full startup of the app.

pmacius (approved):

charts/alfresco-activiti/README.md
```diff
@@ -64,11 +64,12 @@ A Helm chart for Alfresco Activiti
...
-| livenessProbe.failureThreshold | int | `5` |  |
-| livenessProbe.initialDelaySeconds | int | `25` |  |
-| livenessProbe.path | string | `"/activiti-app/app/rest/locale"` |  |
-| livenessProbe.periodSeconds | int | `10` |  |
-| livenessProbe.timeoutSeconds | int | `5` |  |
+| livenessProbe.<<.failureThreshold | int | `5` |  |
+| livenessProbe.<<.initialDelaySeconds | int | `25` |  |
+| livenessProbe.<<.path | string | `"/activiti-app/actuator/health/readiness"` | Make sure to set `readinessProbe.path` to `/activiti-app/app/rest/locale` if you're using a version of APS prior to 24.3.0 |
+| livenessProbe.<<.periodSeconds | int | `10` |  |
+| livenessProbe.<<.timeoutSeconds | int | `5` |  |
```
gionn, line 71: helm-docs doesn't seems to like that syntax, I prefer we just repeat those few keys

alxgomz: mmmm Indeed, I missed that

charts/alfresco-activiti/values.yaml
```diff
@@ -60,13 +60,20 @@ persistence:
...
-  path: /activiti-app/app/rest/locale
+readinessProbe:
+  # -- Make sure to set `readinessProbe.path` to `/activiti-app/app/rest/locale` if you're using a version of APS prior to 24.3.0
+  path: /activiti-app/actuator/health/readiness
+  initialDelaySeconds: 25
+  periodSeconds: 10
+  failureThreshold: 5
+  timeoutSeconds: 5
+livenessProbe:
+  # -- Make sure to set `lvenessProbe.path` to `/activiti-app/app/rest/locale` if you're using a version of APS prior to 24.3.0
```
gionn, line 71:
```suggestion
  # -- For APS prior to 24.3.0, path must be set to `/activiti-app/app/rest/locale`
```

charts/alfresco-activiti/values.yaml
```diff
@@ -60,13 +60,20 @@ persistence:
   data:
     mountPath: "/usr/local/data"
     subPath: "alfresco-process-services/process-data"
-readinessProbe: &apsProbe
-  path: /activiti-app/app/rest/locale
+readinessProbe:
+  # -- Make sure to set `readinessProbe.path` to `/activiti-app/app/rest/locale` if you're using a version of APS prior to 24.3.0
```
gionn, line 64:
```suggestion
  # -- For APS prior to 24.3.0, path must be set to `/activiti-app/app/rest/locale`
```

PR #328, opened 2024-07-10
OPSEXP-2740: bump Alfresco components 

alxgomz:
Ref: OPSEXP-2740

### DIFF
--- a/charts/activemq/Chart.yaml
+++ b/charts/activemq/Chart.yaml
@@ -11,6 +11,6 @@ version: 3.5.3
 appVersion: 5.18.4
 dependencies:
   - name: alfresco-common
-    version: 3.1.2
+    version: 3.1.3
     repository: https://alfresco.github.io/alfresco-helm-charts/
 icon: https://avatars0.githubusercontent.com/u/391127?s=200&v=4
--- a/charts/activemq/Chart.yaml
+++ b/charts/activemq/Chart.yaml
@@ -7,7 +7,7 @@ keywords:
 name: activemq
 sources:
   - https://github.com/Alfresco/alfresco-helm-charts
-version: 3.5.3
+version: 3.6.0
 appVersion: 5.18.4
 dependencies:
   - name: alfresco-common
--- a/charts/activemq/Chart.yaml
+++ b/charts/activemq/Chart.yaml
@@ -7,7 +7,7 @@ keywords:
 name: activemq
 sources:
   - https://github.com/Alfresco/alfresco-helm-charts
-version: 3.6.0
+version: 3.5.4
 appVersion: 5.18.4
 dependencies:
   - name: alfresco-common
--- a/charts/alfresco-activiti-admin/Chart.yaml
+++ b/charts/alfresco-activiti-admin/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v2
 description: A Helm chart for Alfresco Activiti
 name: alfresco-activiti-admin
-version: 0.1.1
+version: 0.2.0
 appVersion: 24.2.1
 dependencies:
   - name: alfresco-common
--- a/charts/alfresco-activiti-admin/Chart.yaml
+++ b/charts/alfresco-activiti-admin/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v2
 description: A Helm chart for Alfresco Activiti
 name: alfresco-activiti-admin
-version: 0.2.0
+version: 0.1.2
 appVersion: 24.2.1
 dependencies:
   - name: alfresco-common
--- a/charts/alfresco-activiti-admin/Chart.yaml
+++ b/charts/alfresco-activiti-admin/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 description: A Helm chart for Alfresco Activiti
 name: alfresco-activiti-admin
 version: 0.1.1
-appVersion: 24.2.0
+appVersion: 24.2.1
 dependencies:
   - name: alfresco-common
     version: 3.1.2
--- a/charts/alfresco-activiti-admin/Chart.yaml
+++ b/charts/alfresco-activiti-admin/Chart.yaml
@@ -5,7 +5,7 @@ version: 0.1.1
 appVersion: 24.2.1
 dependencies:
   - name: alfresco-common
-    version: 3.1.2
+    version: 3.1.3
     repository: https://alfresco.github.io/alfresco-helm-charts/
   - name: postgresql
     repository: oci://registry-1.docker.io/bitnamicharts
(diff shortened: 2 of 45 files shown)


charts/activemq/Chart.yaml
```diff
@@ -7,10 +7,10 @@ keywords:
 name: activemq
 sources:
   - https://github.com/Alfresco/alfresco-helm-charts
-version: 3.5.3
+version: 3.6.0
```
pmacius, line 10: Maybe we can bump to 3.5.4 since there is only a patch bump in dependencies

charts/alfresco-activiti-admin/Chart.yaml
```diff
@@ -1,11 +1,11 @@
 apiVersion: v2
 description: A Helm chart for Alfresco Activiti
 name: alfresco-activiti-admin
-version: 0.1.1
-appVersion: 24.2.0
+version: 0.2.0
```
gionn, line 4:
```suggestion
version: 0.1.2
```

all the bumps are bugfixes

charts/alfresco-activiti/Chart.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: v2
 description: A Helm chart for Alfresco Activiti
 name: alfresco-activiti
-version: 0.2.0-alpha.1
+version: 0.2.0-alpha.2
 appVersion: 24.3.0-SNAPSHOT
```
gionn, line 5: isn't automated?

charts/alfresco-activiti/README.md
```diff
@@ -55,7 +55,7 @@ A Helm chart for Alfresco Activiti
 | image.internalPort | int | `8080` |  |
 | image.pullPolicy | string | `"IfNotPresent"` |  |
 | image.repository | string | `"quay.io/alfresco/alfresco-process-services"` |  |
-| image.tag | string | `"develop-11791"` |  |
+| image.tag | string | `"develop-12323"` |  |
```
gionn, line 58: now we should use GA here

charts/alfresco-adf-app/Chart.yaml
```diff
@@ -2,8 +2,8 @@ apiVersion: v2
 name: alfresco-adf-app
 description: A generic Alfresco Development Framework Helm chart for Kubernetes
 type: application
-version: 0.1.0
+version: 0.2.0
```
gionn, line 5:
```suggestion
version: 0.1.1
```

charts/alfresco-search-enterprise/Chart.yaml
```diff
@@ -3,7 +3,7 @@ apiVersion: v2
 name: alfresco-search-enterprise
 description: A Helm chart for deploying Alfresco Elasticsearch connector
 type: application
-version: 4.0.0-alpha.0
+version: 4.0.0
```
gionn, line 6: if there aren't any changes there is no need to release now

charts/alfresco-search-service/Chart.yaml
```diff
@@ -10,12 +10,12 @@ keywords:
 name: alfresco-search-service
 sources:
   - https://github.com/Alfresco/alfresco-helm-charts
-version: 3.3.1
-appVersion: 2.0.10
+version: 3.4.0
```
gionn, line 13:
```suggestion
version: 3.3.2
```

charts/alfresco-search-service/charts/alfresco-insight-zeppelin/Chart.yaml
```diff
@@ -11,8 +11,8 @@ name: alfresco-insight-zeppelin
 version: 3.0.1
```
gionn, line 11: I don't know what we should do with this (keep in sync with parent?)

charts/alfresco-sync-service/values.yaml
```diff
@@ -17,7 +17,7 @@ podSecurityContext:
   fsGroup: 1000
 image:
   repository: quay.io/alfresco/service-sync
-  tag: 5.0.0
+  tag: 4.0.1
```
gionn, line 20: should stay on 5.x since it's latest GA

gionn (changes requested):

charts/activemq/Chart.yaml
alxgomz: Guys that does not comply to the defined versioning scheme: https://hyland.atlassian.net/wiki/spaces/TECH/pages/1293317631/Charts+release+versioning+scheme
I'm fine with changing it but I think it's time to review and approve (👍) that doc (it's still in "ready for review")

charts/alfresco-activiti/Chart.yaml
alxgomz: It is but I reverted the change as the current code require unreleased 24.3.0 (and it's not used anywhere at that point)

charts/alfresco-activiti/README.md
alxgomz: same as aboce

charts/alfresco-activiti/Chart.yaml
alxgomz: I will simply not release anything here for now

charts/alfresco-activiti-admin/Chart.yaml
alxgomz: same as above

charts/alfresco-sync-service/values.yaml
alxgomz: good point, the 4.x enforcement for acs 23.2 needs to happen in the umbrella chart 👍

charts/activemq/Chart.yaml
gionn: looks more an uncovered case rather than not complying, what is the specific point you are referring to?

also we need to have this moved or at least mentioned in the developers/release md

charts/alfresco-sync-service/values.yaml
gionn: I remember reverting this manually last time we did the release, there is not an easy way to handle it via updatecli matrix

charts/alfresco-search-service/charts/alfresco-insight-zeppelin/Chart.yaml
alxgomz: I will just bump it independently. There very few chance it ever evolves without search service but that's still a different code base where bumps should be independent (e.g. a bug fix in one of the chart themselves)

charts/activemq/Chart.yaml
alxgomz: the page is indeed unclear but I'm prety sure we decided we would decoralate bugfixes from images and dependencies from actual chart bugfixes... Though thinking about it now seems overly complicated and treating all bugfixes the same is much easier. Hence, I updated the page and linked in the Release section.

pmacius (approved):

charts/alfresco-activiti/Chart.yaml
gionn: ah ok, I was hoping the fix would be in the GA we are currently releasing

gionn (approved):

PR #343, opened 2024-08-19
OPSEXP-2785 Add completely customizable probes for connector charts

pmacius:
Ref: OPSEXP-2785

### DIFF
--- a/charts/alfresco-connector-ms365/Chart.yaml
+++ b/charts/alfresco-connector-ms365/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: alfresco-connector-ms365
 description: A Helm chart for deploying Alfresco connector ms365 service
 type: application
-version: 1.0.0
+version: 2.0.0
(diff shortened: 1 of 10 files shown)


gionn (changes requested):
drop image.internalPort from default values if not used elsewhere (spoiler: it was)

gionn (approved):

pmacius:
@alxgomz I guess its safer to keep it major; when some user is defining different .internalPort it might be an issue.

charts/alfresco-connector-ms365/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v2
 name: alfresco-connector-ms365
 description: A Helm chart for deploying Alfresco connector ms365 service
 type: application
-version: 1.0.0
+version: 2.0.0
```
gionn, line 5: release an alpha for now, will do a GA with the image bumps next week

gionn (approved):

PR #350, opened 2024-08-26
OPSEXP-2778 Bump tags for Indus release

pmacius:
Ref: OPSEXP-2778

### DIFF
--- a/charts/activemq/Chart.yaml
+++ b/charts/activemq/Chart.yaml
@@ -7,7 +7,7 @@ keywords:
 name: activemq
 sources:
   - https://github.com/Alfresco/alfresco-helm-charts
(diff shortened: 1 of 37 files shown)


charts/activemq/Chart.yaml
```diff
@@ -7,8 +7,8 @@ keywords:
 name: activemq
 sources:
   - https://github.com/Alfresco/alfresco-helm-charts
-version: 3.5.4
-appVersion: 5.18.4
+version: 3.5.5
```
gionn, line 10: it's better if you release first activemq alone so other charts could pick this version

gionn (changes requested):

gionn (approved):

PR #365, opened 2024-09-17
Fix csrf.origin url join for share-config

swagnerfaw:
According to the share-config.xml CSRF origins should be separated with  | and not ,


gionn (approved):

gionn:
Thanks @swagnerfaw! 🚀

PR #375, opened 2024-09-25
OPSEXP-2482 Replace deprecated annotations with spec.ingressClassName

pmacius:
ref: OPSEXP-2482


charts/alfresco-connector-ms365/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v2
 name: alfresco-connector-ms365
 description: A Helm chart for deploying Alfresco connector ms365 service
 type: application
-version: 2.0.0
+version: 2.0.1
```
gionn, line 5: should be worth a minor (if not a major) try to find what we did for other charts with this change

gionn (approved):

PR #376, opened 2024-09-25
OPSEXP-2654 Reindex job as configurable helm hook

gionn:
OPSEXP-2654

The new default is to run it as post-install hook so it won't rerun on each upgrade if was cleaned up.

Optionally `reindexing.hookExecution` could be set to `post-upgrade` so that could mimic an on-demand execution by triggering together with an upgrade. In this case the job name got a rand appended to avoid clashes.

It can be set to false to restore the previous behavior.

### DIFF
--- a/charts/alfresco-search-enterprise/ci/default-values.yaml
+++ b/charts/alfresco-search-enterprise/ci/default-values.yaml
@@ -4,7 +4,7 @@ reindexing:
   # a db would be required but helm install won't wait for the job to be ready
   enabled: true
   # ... but only if it's not defined as an hook
-  hookExecution: post-install
+  hookExecution: false
   repository:
     url: http://repository
   db:
--- a/charts/alfresco-search-enterprise/ci/default-values.yaml
+++ b/charts/alfresco-search-enterprise/ci/default-values.yaml
@@ -1,8 +1,10 @@
 # avoid too long resource names being truncated and conflicting
 nameOverride: aesc
 reindexing:
-  # requires postgres database at startup
+  # a db would be required but helm install won't wait for the job to be ready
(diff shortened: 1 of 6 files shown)


charts/alfresco-search-enterprise/ci/default-values.yaml
```diff
@@ -1,8 +1,10 @@
 # avoid too long resource names being truncated and conflicting
 nameOverride: aesc
 reindexing:
-  # requires postgres database at startup
+  # a db would be required but helm install won't wait for the job to be ready
   enabled: true
+  # ... but only if it's not defined as an hook
```
pmacius, line 6: Im not sure if I understand the comments: So the db is not required with job defined as a post-install hook?

charts/alfresco-search-enterprise/values.yaml
```diff
@@ -187,6 +187,9 @@ reindexing:
     pullPolicy: IfNotPresent
   pathIndexingEnabled: true
   environment: {}
+  # -- When to execute the reindexing job (e.g. `post-install` or `post-upgrade`
+  # or set to false to apply as a standard resource)
+  hookExecution: post-install
```
pmacius, line 192: I would also add a possibility to define https://helm.sh/docs/topics/charts_hooks/#hook-deletion-policies

gionn: I preferred using `ttlSecondsAfterFinished` which applies anyway (even when not using hooks) and give a configurable amount of time to review logs, regardless of the outcome

pmacius (approved):

alxgomz (approved):

PR #387, opened 2024-10-07
OPSEXP-2882 Add json configmap configurable with yaml to adf chart

pmacius:
ref: OPSEXP-2882

### DIFF
--- a/charts/alfresco-adf-app/templates/configmap.yaml
+++ b/charts/alfresco-adf-app/templates/configmap.yaml
@@ -1,0 +1,11 @@
+{{- if .Values.configMap }}
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: {{ include "alfresco-adf-app.fullname" . }}
+  labels:
+    {{- include "alfresco-adf-app.labels" . | nindent 4 }}
+data:
+  config.json: >-
+    {{- .Values.configMap | toJson | nindent 4 }}
+{{- end }}
--- a/charts/alfresco-adf-app/templates/configmap.yaml
+++ b/charts/alfresco-adf-app/templates/configmap.yaml
@@ -6,6 +6,6 @@ metadata:
   labels:
     {{- include "alfresco-adf-app.labels" . | nindent 4 }}
 data:
-  config.json: >-
+  app.config.json: >-
     {{- .Values.configMap | toJson | nindent 4 }}
 {{- end }}
--- a/charts/alfresco-adf-app/values.yaml
+++ b/charts/alfresco-adf-app/values.yaml
@@ -78,7 +78,7 @@ volumes: []
 # -- Additional volumeMounts on the output Deployment definition
 volumeMounts: []
 
-configMap: []
+configMap: {}
 
 nodeSelector: {}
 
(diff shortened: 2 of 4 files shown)


charts/alfresco-adf-app/templates/configmap.yaml
```diff
@@ -0,0 +1,11 @@
+{{- if .Values.configMap }}
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: {{ include "alfresco-adf-app.fullname" . }}
+  labels:
+    {{- include "alfresco-adf-app.labels" . | nindent 4 }}
+data:
+  config.json: >-
```
pmacius, line 9: not sure if we can hardcode that and if we need to add the volume mounts but I think its better if we leave volume mounts optional

alxgomz: not only can we but we have to. However it should be named `app.config.json` IIRC

charts/alfresco-adf-app/values.yaml
```diff
@@ -78,6 +78,8 @@ volumes: []
 # -- Additional volumeMounts on the output Deployment definition
 volumeMounts: []
 
+configMap: []
```
alxgomz, line 81: If configmap is a list then I'm not sure how this works.
Do you expect to render a json list (I think that's not the expected format)? or do you plan to have more than one confimap (in which case we'd need an iterator somewhere)?

alxgomz (changes requested):

charts/alfresco-adf-app/values.yaml
pmacius: I was planning just to have one configmap. Whatever you put in yaml format it will be converted to json. I our case it can look like this:
```
configMap:
  $schema: "../../../node_modules/@alfresco/adf-core/app.config.schema.json"
  providers: ALL
  ecmHost: "{protocol}//{hostname}{:port}"
  bpmHost: "{protocol}//{hostname}{:port}"
  authType: OAUTH
  ...
``` 
Let me change that to {}

charts/alfresco-adf-app/values.yaml
```diff
@@ -78,6 +78,8 @@ volumes: []
 # -- Additional volumeMounts on the output Deployment definition
 volumeMounts: []
 
+configMap: {}
```
alxgomz, line 81: Can you add an helm-docs annotation so users know what this is all about?

alxgomz (changes requested):

charts/alfresco-adf-app/templates/configmap.yaml
```diff
@@ -0,0 +1,11 @@
+{{- if .Values.configMap }}
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: {{ include "alfresco-adf-app.fullname" . }}
```
pmacius, line 5: I am thinking that maybe we can hardcode the name of that configmap to make it easier for users to mount it. Not sure

alxgomz: If user wants to mount his own cm I think we agreed he would:
* create his own cm
* create an extra volume for it (where he can use whatever name he wants)
* create the volume mount that goes with it
and so we would not have this configMap at all

If we agree on that, then we should document it in this chart

PR #393, opened 2024-10-11
OPSEXP-2877 Add alfresco-audit-storage chart

pmacius:
Ref: OPSEXP-2877


charts/alfresco-audit-storage/values.yaml
```diff
@@ -0,0 +1,127 @@
...
+  initialDelaySeconds: 15
+  periodSeconds: 60
+readinessProbe:
+  httpGet:
+    path: /actuator/health/readiness
+    port: http
+  initialDelaySeconds: 15
+  periodSeconds: 60
+
+search:
```
gionn, line 57: index?

this has really not something to do with searching capabilities

charts/alfresco-audit-storage/templates/deployment.yaml
```diff
@@ -0,0 +1,74 @@
...
+  selector:
+    matchLabels:
+      {{- include "alfresco-audit-storage.selectorLabels" . | nindent 6 }}
+  template:
+    metadata:
+      annotations:
+        checksum/config-es: {{ include (print $.Template.BasePath "/configmap-es.yaml") $ | sha256sum }}
+        checksum/config-mq: {{ include (print $.Template.BasePath "/configmap-mq.yaml") $ | sha256sum }}
+        checksum/secret-es: {{ include (print $.Template.BasePath "/secret-es.yaml") $ | sha256sum }}
+        checksum/secret-mq: {{ include (print $.Template.BasePath "/secret-mq.yaml") $ | sha256sum }}
```
alxgomz, line 18: I would not use that as it do not fit with the approach we use widely to offer usage of exisitingConfigmap (and has other drawbacks). We've been removing it from other charts (although some are probably still around). I think we should have a consistent recommendation of using proper tools like [reloader](https://github.com/stakater/Reloader)

charts/alfresco-audit-storage/values.yaml
alxgomz: Yeah worth renaming as it gives the impression it's tied the the full text search index as it is.

charts/alfresco-audit-storage/templates/deployment.yaml
gionn: I think we never took a final decision on this, for me they are still worth to have because **sometimes** they help the end users without requiring any external tooling, so dropping them mean making life harder **every time** for end users.

gionn: service restarts should not really be an issue on a kubernetes environment, plus this approach is also documented on helm docs https://helm.sh/docs/howto/charts_tips_and_tricks/#automatically-roll-deployments

however not a big deal for me if you prefer to get rid of it

charts/alfresco-audit-storage/Chart.yaml
```diff
@@ -0,0 +1,20 @@
+apiVersion: v2
+name: alfresco-audit-storage
+description: A Helm chart for Kubernetes to deploy Alfresco Audit Storage
+type: application
+version: 0.0.1-alpha.0
```
gionn, line 5:
```suggestion
version: 0.1.0-alpha.0
```

the first alpha can't be a bugfix release - anyway not a big deal

gionn (approved):

PR #400, opened 2024-10-28
OPSEXP-2885 Fix secret template bug in search enterprise correcting search username reference

pmacius:
Ref: OPSEXP-2885


gionn (changes requested):
write a unit for the secret which should have exposed this issue before

pmacius:
> write a unit for the secret which should have exposed this issue before

Test for creating this secret are already there but this scenario Im not sure can be somehow covered.

gionn (approved):

PR #423, opened 2024-12-06
OPSEXP-2933 Point links to support.hyland.com

pmacius:
Ref: OPSEXP-2933


charts/alfresco-repository/values.yaml
pmacius, line 1: not sure if worth to bump chart version for this change

pmacius: ah yes linter says yes

alxgomz (approved):

PR #425, opened 2024-12-10
OPSEXP-2984 Remove dependency to bitnami/common in alfresco-common chart

pmacius:
Ref: OPSEXP-2984

### DIFF
--- a/charts/alfresco-common/Chart.yaml
+++ b/charts/alfresco-common/Chart.yaml
@@ -5,5 +5,5 @@ description: |
   A helper subchart to avoid duplication in alfresco charts and set common
   external dependencies
 type: library
-version: 4.0.0
+version: 4.0.0-alpha.0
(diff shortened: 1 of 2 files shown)


charts/alfresco-common/Chart.yaml
```diff
@@ -5,9 +5,5 @@ description: |
   A helper subchart to avoid duplication in alfresco charts and set common
   external dependencies
 type: library
-version: 3.1.4
-dependencies:
-  - name: common
-    repository: oci://registry-1.docker.io/bitnamicharts
-    version: 1.x.x
+version: 4.0.0
```
gionn, line 8:
```suggestion
version: 4.0.0-alpha.0
```

given that we need to test every other chart before

gionn (changes requested):

gionn (approved):

PR #430, opened 2024-12-20
OPSEXP-2991 Change sync service ingress pathType to ImplementationSpecific

pmacius:
Ref: OPSEXP-2991

### DIFF
--- a/charts/alfresco-sync-service/README.md
+++ b/charts/alfresco-sync-service/README.md
@@ -47,7 +47,11 @@ Checkout [alfresco-content-services chart's doc](https://github.com/Alfresco/acs
 | image.pullPolicy | string | `"IfNotPresent"` |  |
 | image.repository | string | `"quay.io/alfresco/service-sync"` |  |
 | image.tag | string | `"5.1.0"` |  |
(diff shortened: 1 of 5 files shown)


charts/alfresco-sync-service/README.md
```diff
@@ -49,7 +49,7 @@ Checkout [alfresco-content-services chart's doc](https://github.com/Alfresco/acs
 | image.tag | string | `"5.1.0"` |  |
 | ingress.extraAnnotations | object | `{}` | useful when running Sync service without SSL termination done by a load balancer, e.g. when ran on Minikube for testing purposes nginx.ingress.kubernetes.io/ssl-redirect: "false" |
 | ingress.hosts[0].paths[0].path | string | `"/syncservice(/|$)(.*)"` |  |
-| ingress.hosts[0].paths[0].pathType | string | `"Prefix"` |  |
+| ingress.hosts[0].paths[0].pathType | string | `"ImplementationSpecific"` |  |
```
gionn, line 52: is really necessary to keep using this regexp? I remember it's here since forever and I can't understand which a simple `Prefix` & `/syncservice` should not have the same behavior

gionn (changes requested):

charts/alfresco-sync-service/README.md
pmacius: As we discussed ingress is using rewrite target so it is required do have this regex: 
https://kubernetes.github.io/ingress-nginx/examples/rewrite/#rewrite-target

gionn (approved):

PR #434, opened 2025-01-20
OPSEXP-2249 Enable helm unittest with subcharts

pmacius:
Ref: OPSEXP-2249


alxgomz (changes requested):
set a test value file in the zeppelin test folder and use `values` in the test definition ( as in https://github.com/Alfresco/alfresco-helm-charts/blob/fd9dac41efe18e059cc8cd26bd0127e83fdfd466/charts/alfresco-repository/tests/deployment_test.yaml#L6)

PR #441, opened 2025-01-29
OPSEXP-3034 Bump charts for Pegasus release

pmacius:
ref: OPSEXP-3034


gionn (changes requested):
we want to bundle in this release also the values update (otherwise you will have to release twice each chart).

just run the updatecli workflow against this branch (review updatecli supported matrix first)

gionn (approved):

PR #444, opened 2025-02-03
OPSEXP-3028: bump KinD to reflect EKS supported versions 

alxgomz:
Ref: OPSEXP-3028

### DIFF
--- a/.github/workflows/lint-test.yaml
+++ b/.github/workflows/lint-test.yaml
@@ -22,8 +22,8 @@ jobs:
     env:
       INSTALL_NAMESPACE: alfresco
       # See https://github.com/kubernetes-sigs/kind/releases for the latest available images
-      KIND_VERSION: v0.22.0
-      KIND_NODE_IMAGE: kindest/node:v1.29.2@sha256:51a1434a5397193442f0be2a297b488b6c919ce8a3931be0ce822606ea5ca245
+      KIND_VERSION: v0.26.0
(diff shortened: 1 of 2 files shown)


pmacius (approved):

.github/workflows/lint-test.yaml
```diff
@@ -22,8 +22,8 @@ jobs:
     env:
       INSTALL_NAMESPACE: alfresco
       # See https://github.com/kubernetes-sigs/kind/releases for the latest available images
-      KIND_VERSION: v0.22.0
-      KIND_NODE_IMAGE: kindest/node:v1.29.2@sha256:51a1434a5397193442f0be2a297b488b6c919ce8a3931be0ce822606ea5ca245
+      KIND_VERSION: v0.26.0
+      KIND_NODE_IMAGE: kindest/node:v1.31.4@sha256:2cb39f7295fe7eafee0842b1052a599a4fb0f8bcf3f83d96c7f4864c357c6c30
```
gionn, line 26: I remember discussing it with Pawel but don't remember the reason - why we are not using the build-tools setup-kind here? at least it should allow us to bump version in a single place

alxgomz: I didn't even pay attention tbh...
That said I would still be specific about the versions we want here and avoid using defaults. In my mind the build tools actions should be kept up to date on bleeding edge and consumer shoudl pin versions of kind, nodes and nginx... Well wasn't nginx the reason for not using the build tools action?

gionn (approved):

PR #451, opened 2025-03-05
OPSEXP-3050 Move server snippet annotation templates to componets charts

pmacius:
Ref: OPSEXP-3050

### DIFF
--- a/charts/alfresco-repository/templates/ingress.yaml
+++ b/charts/alfresco-repository/templates/ingress.yaml
@@ -13,7 +13,7 @@ metadata:
   annotations:
     checkov.io/skip1: CKV_K8S_153=We're filtering out snippet in named template
     {{- include "alfresco-common.nginx.annotations" .Values }}
-    {{- include "alfresco-common.nginx.secure.annotations" .Values }}
+    {{- include "alfresco-repository.nginx.secure.annotations" . | nindent 4 }}
 spec:
   ingressClassName: {{ .Values.ingress.className }}
   {{- if .Values.ingress.tls }}
--- a/charts/alfresco-share/tests/ingress_test.yaml
+++ b/charts/alfresco-share/tests/ingress_test.yaml
@@ -3,18 +3,6 @@ suite: test share ingress
 templates:
   - ingress.yaml
 tests:
-  - it: should render with default security annotations
-    asserts:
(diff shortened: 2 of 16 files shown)


charts/alfresco-share/values.yaml
```diff
@@ -70,6 +70,8 @@ ingress:
     nginx.ingress.kubernetes.io/session-cookie-path: "/share"
     nginx.ingress.kubernetes.io/session-cookie-max-age: "604800"
     nginx.ingress.kubernetes.io/session-cookie-expires: "604800"
+  serverSnippet:
+    solrEnabled: false
```
gionn, line 74: better have it enabled by default to avoid forgetting to set it

charts/alfresco-share/tests/ingress_test.yaml
```diff
@@ -8,10 +8,7 @@ tests:
       - equal:
           path: metadata.annotations['nginx.ingress.kubernetes.io/server-snippet']
           value: |
-            location ~ ^/.*/(wc)?s(ervice)?/api/solr/.*$ {return 403;}
-            location ~ ^/.*/proxy/.*/api/solr/.*$ {return 403;}
             location ~ ^/.*/-default-/proxy/.*/api/.*$ {return 403;}
-            location ~ ^/.*/s/prometheus$ {return 403;}
```
gionn, line 14: set both cases

charts/alfresco-repository/templates/ingress.yaml
```diff
@@ -13,7 +13,7 @@ metadata:
   annotations:
     checkov.io/skip1: CKV_K8S_153=We're filtering out snippet in named template
     {{- include "alfresco-common.nginx.annotations" .Values }}
-    {{- include "alfresco-common.nginx.secure.annotations" .Values }}
+    {{- include "alfresco-repository.nginx.secure.annotations" . | nindent 4 }}
```
gionn, line 16: I think it needs to be moved before the user annotations or users can't never override it, make sure to test it as well

gionn (changes requested):

pmacius:
server-snippet annotations on ingress-nginx can be applied only once per host (ingress-controller pod) so having multiple definition of it depending on the component is not possible

PR #452, opened 2025-03-06
OPSEXP-3128 Fixup compatiblity with nginx-unprivileged images

gionn:
Bumping charts in https://github.com/Alfresco/alfresco-dockerfiles-bakery/pull/153 exposed this breaking change

uid/gid should match the ones of the upstream `nginxinc/nginx-unprivileged` image so the entrypoint customization script works as expected.

OPSEXP-3128

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -19,7 +19,7 @@ repos:
     hooks:
       - id: kubeconform-helm
         alias: kubeconform-helm-min
-        name: "Kubeconform Helm - min k8s version" # most older EKS version still in extended support
+        name: "Kubeconform Helm - min k8s version" # the oldest EKS version still in extended support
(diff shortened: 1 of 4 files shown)


.pre-commit-config.yaml
```diff
@@ -19,7 +19,7 @@ repos:
     hooks:
       - id: kubeconform-helm
         alias: kubeconform-helm-min
-        name: "Kubeconform Helm - min k8s version (most recent EKS version inextended support)"
+        name: "Kubeconform Helm - min k8s version" # most older EKS version still in extended support
```
pmacius, line 22:
```suggestion
        name: "Kubeconform Helm - min k8s version" # the oldest EKS version still in extended support
```

pmacius (approved):

PR #459, opened 2025-03-29
alfresco-repository: conditionally remove rollingUpdate block when strategy.type=Recreate

Mesh33:
### Summary
This PR fixes a deployment failure when `strategy.type` is set to `Recreate`. Previously, the chart always included the `rollingUpdate` configuration, causing an error in Kubernetes (`spec.strategy.rollingUpdate` is invalid when `type: Recreate`) -> is invalid: spec.strategy.rollingUpdate: Forbidden: may not be specified when strategy `type` is `Recreate`

### What has changed
- The `strategy` block in `alfresco-repository` now conditionally includes the `rollingUpdate` section **only** if `strategy.type` (or the global fallback) is `"RollingUpdate"`.
- The implementation uses `coalesce` and a `default` to ensure a valid fallback strategy is always set.

### Why this is needed
Without this fix, users overriding the deployment strategy to `Recreate` face errors during deployment due to invalid manifest generation. This PR ensures that Helm only emits `rollingUpdate` fields if they are actually required.
I have exhaustively tested all possible alternatives, including:
- Setting `rollingUpdate` to `null`
- Setting `rollingUpdate` to `{}` (empty object)
- Setting `maxSurge` and `maxUnavailable` to `null`
- Explicitly removing `rollingUpdate` in values

### Testing
- Rendered the template locally via `helm template` to confirm that `rollingUpdate` is omitted when `strategy.type=Recreate`.
- Verified backward compatibility to ensure the chart defaults to `RollingUpdate` when no strategy is defined.

Thanks

### DIFF
--- a/charts/alfresco-repository/templates/deployment.yaml
+++ b/charts/alfresco-repository/templates/deployment.yaml
@@ -12,7 +12,11 @@ spec:
     matchLabels:
       {{- include "alfresco-repository.selectorLabels" . | nindent 6 }}
   strategy:
-    {{- toYaml (.Values.strategy | default .Values.global.strategy) | nindent 4 }}
+    type: {{ coalesce .Values.strategy.type .Values.global.strategy.type | default "RollingUpdate" }}
+    {{- if eq (coalesce .Values.strategy.type .Values.global.strategy.type | default "RollingUpdate") "RollingUpdate" }}
+    rollingUpdate:
+      {{- toYaml (coalesce .Values.strategy.rollingUpdate .Values.global.strategy.rollingUpdate) | nindent 6 }}
+    {{- end }}
   template:
     metadata:
       annotations:


charts/alfresco-repository/templates/deployment.yaml
```diff
@@ -12,7 +12,11 @@ spec:
     matchLabels:
       {{- include "alfresco-repository.selectorLabels" . | nindent 6 }}
   strategy:
+    {{- if eq (.Values.strategy.type | default .Values.global.strategy.type) "Recreate" }}
+    type: Recreate
+    {{- else }}
     {{- toYaml (.Values.strategy | default .Values.global.strategy) | nindent 4 }}
+    {{- end }}
```
gionn, line 19: Hello, thanks for raising this contribution!

I think the approach to solving this could be made more explicit with (although untested):

```suggestion
    type: {{ .Values.strategy.type | default .Values.global.strategy.type }}
    {{- if eq (.Values.strategy.type | default .Values.global.strategy.type) "RollingUpdate" }}
    rollingUpdate: {{- toYaml (.Values.strategy.rollingUpdate | default .Values.global.strategy.rollingUpdate) | nindent 6 }}
    {{- end }}
```

in this way should be clear that we are preventing to have the `rollingUpdate` key when strategy is not `RollingUpdate`, and is similar to what we are already doing in another chart https://github.com/Alfresco/alfresco-helm-charts/blob/48a7d36c068c9401f180176474df1fb9b68918e9/charts/alfresco-transform-service/templates/deployment-filestore.yaml#L15

Finally the best would be to improve the unit tests to cover this aspect, by adding an additional assert for the default manifest test:
https://github.com/Alfresco/alfresco-helm-charts/blob/main/charts/alfresco-repository/tests/deployment_test.yaml#L8

and eventually add a new scenario at the end of the same fail asserting that in case `.Values.strategy.type: Recreate` we don't get the additional section.

gionn (changes requested):

charts/alfresco-repository/templates/deployment.yaml
Mesh33: Hi @gionn ,

Thanks for your feedback.

I updated the template and added the missing unit tests.

Mesh33: Hey @gionn ,
I've added changes to support **global.strategy**. However, because this chart already sets **strategy** by default in **alfresco-repository/values.yaml**, I'm not entirely convinced there's a practical need to handle **global.strategy**, maybe when values.strategy is forced to null ? (I added an unit test)
To avoid forcing defaults on the global object in alfresco-repository values.yaml (because global: has sense only in the main helm chart?), I had to add logic for when global.strategy is absent.

Let me know your thoughts. It might be simpler to stick to Values.strategy alone, but I'm open to any recommendations you have.

Thanks,

gionn: > I'm not entirely convinced there's a practical need to handle **global.strategy**

Looks like it's coming from a 6 years old PR:

https://github.com/Alfresco/acs-deployment/commit/09b41c0c2eddad5985238ceb931d564d881ff1c5

I think the intention was to provide a way to change strategy on multiple deployments setting only a single value, but I am more inclined to simplify it rather than making it more complex to handle a specific use case.

You can get rid of `.Values.global.strategy.type` usage for me, having a default in values is already fine.

gionn (approved):

Issue #463, opened 2025-04-16
no possibility to mount truststore for alfresco-search-enterprise


wimfabri:
Hello,

the alfresco-search-enterprise does not provide a way to add a truststore for elasticsearch. We have elasticsearch running with a self-signed certificate for https so this is needed.

alxgomz:
Thank you for reporting this issue.
We recognise this is missing and have added this to our backlog.
In the mean time a workaround could be to build the image directly with the updated truststore shipped inside the image.
That process should be more simple now with [Alfresco Bakery](https://github.com/Alfresco/alfresco-dockerfiles-bakery)

PR #470, opened 2025-05-06
OPSEXP-3225: allow mounting extra volumes in search enterprise

alxgomz:
Ref: OPSEXP-3225

Provides comprehensive instructions on how to mount a custom keystore (addresses #463 )

### DIFF
--- a/charts/alfresco-search-enterprise/Chart.yaml
+++ b/charts/alfresco-search-enterprise/Chart.yaml
@@ -3,7 +3,7 @@ apiVersion: v2
 name: alfresco-search-enterprise
 description: A Helm chart for deploying Alfresco Elasticsearch connector
 type: application
-version: 4.4.0
+version: 4.4.1
 appVersion: 5.0.0
 dependencies:
   - name: alfresco-common
--- a/charts/alfresco-search-enterprise/docs/truststore.md
+++ b/charts/alfresco-search-enterprise/docs/truststore.md
@@ -1,0 +1,81 @@
+# Setting custom java truststore
+
+There multiple use cases where you may want to set a custom keystore or
+truststore. It can be because you need to interact with a third party service
+which uses a self-signed certificate or a prvite CA that is not shipped with
+the container. Or maybe the service you are trying to connect to uses a
+self-signed certificate and you want to trust it. In this case, you can
+create a custom truststore and set it in the container.
+
+## Create a custom truststore
+
+You can create a custom truststore on your local machine using the
+`keytool` command. The `keytool` command is part of the Java Development Kit (JDK)
+and is used to manage keystores and truststores. You can use the following
+command to create a new truststore and import a certificate into it:
+
+```bash
+keytool -importcert -file <path-to-certificate-file> -alias <alias-name> -keystore <path-to-truststore-file>
+```
+
+Where `<path-to-certificate-file>` is the path to the CA certificate or the
+self-signed certificate you want to trust, `<alias-name>` is a name you want
+to give to the certificate in the truststore, and `<path-to-truststore-file>`
+is the path to the truststore file you want to create.
+You will be prompted to enter a password for the truststore. You can use
+the default password `changeit` or leave it empty (Truststore usually contain
+public informations so it should not be a problem to not use a password).
+Also note that <path-to-truststore-file> can point to an already existing
+truststore file. We recommand using the default  `cacerts` truststore that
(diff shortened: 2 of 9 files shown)


charts/alfresco-search-enterprise/templates/deployment.yaml
```diff
@@ -15,9 +15,6 @@ spec:
   template:
     metadata:
       annotations:
-        checksum/secret-broker: {{ include (print $.Template.BasePath "/secret-messagebroker.yaml") $ | sha256sum }}
-        checksum/secret-database: {{ include (print $.Template.BasePath "/secret-database.yaml") $ | sha256sum }}
-        checksum/secret-elasticsearch: {{ include (print $.Template.BasePath "/secret-elasticsearch.yaml") $ | sha256sum }}
```
gionn, line 20: keep them, we know they are sub-optimal but having them improve user experience for a subset of use cases and no downsides when using external secrets

charts/alfresco-search-enterprise/Chart.yaml
```diff
@@ -3,7 +3,7 @@ apiVersion: v2
 name: alfresco-search-enterprise
 description: A Helm chart for deploying Alfresco Elasticsearch connector
 type: application
-version: 4.4.0
+version: 4.4.1
```
gionn, line 6: more a (missing) feature than a bugfix

charts/alfresco-search-enterprise/docs/truststore.md
```diff
@@ -0,0 +1,87 @@
...
+self-signed certificate you want to trust, `<alias-name>` is a name you want
+to give to the certificate in the truststore, and `<path-to-truststore-file>`
+is the path to the truststore file you want to create.
+You will be prompted to enter a password for the truststore. You can use
+the default password `changeit` or leave it empty (Truststore usually contain
+public informations so it should not be a problem to not use a password).
+Also note that <path-to-truststore-file> can point to an already existing
+truststore file. We recommand using the default  `cacerts` truststore that
+comes with the JDK. This way, you can add your custom certificates to the
+default truststore and use it in your application.
```
gionn, line 37: isn't this something we are telling over and over again - at a different degree depending on how much we are into describing custom trust store creation?

maybe you can extract this in a dedicated page, maybe still in guides which just describe keystore handling (if really there isn't a better piece of external/official documentation which describe this)

charts/alfresco-search-enterprise/templates/statefulset-mediation.yaml
```diff
@@ -55,7 +54,15 @@ spec:
             httpGet:
               path: /actuator/health
               port: http
-          resources: {{- toYaml .Values.resources | nindent 12 }}
+          resources: {{- toYaml $.Values.resources | nindent 12 }}
```
gionn, line 57: is it correct that resources are underneath the root of values?

charts/alfresco-search-enterprise/Chart.yaml
alxgomz: To a certain extend this is a bug. It's a so common use case (in. particular for this components which talks to different 3rd party services) that I decided to make it a bug in both github and JIRA, hence the version

charts/alfresco-search-enterprise/docs/truststore.md
alxgomz: This is what I started doing but unfortunately it quickly appeared to me this chart is kind of special due to the iterative structure of the values file (liveIndexing.*.*) and the merging mechanism that exists for environment variable. The repo also have a similar page that do not work EXACTLY the same but is close.

charts/alfresco-search-enterprise/templates/deployment.yaml
alxgomz: I mostly removed them because they make the tests less DRY as I would be forced to include the secrets tenplate too and the `documentSelector` feature is not flexible enough to support multiple conditions.
Plus, the checksum mecanism is not present in the mediation statefulset so the component would anyway not work after a CM update

charts/alfresco-search-enterprise/templates/statefulset-mediation.yaml
alxgomz: This is how is used to be I just enforced it after I made the `with` statement line 20 span until the end of the template

pmacius (approved):

gionn (approved):

PR #479, opened 2025-06-03
Feature/Additional properties for share

kirananil007:
Currently, there is no defined way in the helm values.yaml to pass in the additional Global properties and secrets for the alfresco share component.

This PR includes changes to pass in the global properties to the alfresco's share deployment.
- Pass in the additional Global properties for Alfresco share deployment.
- Creates a config map for the share alfresco-global.properties
- Attaches the secrets as env for the share deployment spec.

The main values.yaml will looks like:

```
config:
  repository:
    configMapName: &repository_cmName repository
    additionalGlobalProperties:
      solr.http.connection.timeout: 120000
      notification.email.siteinvite.moderatedsites: false                                                                                                                                                  
      extract.assets.job.cron: 0/15 * * * * ?                                                                                                                        
      extract.assets.threads.size: 1                                                                                                                                                                                                                                             
      extract.assets.small.job.cron: 0 0/2 * * * ? 2050
      extract.assets.medium.job.cron: 0 0/2 * * * ? 2050                                                                                                               
      extract.assets.large.job.cron: 0 0/2 * * * ? 2050                                                                                                                           
      extract.assets.batch.size.small: 5                                                                                                                           
      extract.assets.batch.size.medium: 5
      extract.assets.batch.size.large: 5                                                                                                                                                                                
  share:
    configMapName: share
    additionalGlobalProperties:
      solr.http.connection.timeout: 120000
      notification.email.siteinvite.moderatedsites: false                                                                                                                                                  
      extract.assets.job.cron: 0/15 * * * * ?                                                                                                                        
      extract.assets.threads.size: 1                                                                                                                                                                                                                                             
      extract.assets.small.job.cron: 0 0/2 * * * ? 2050
      extract.assets.medium.job.cron: 0 0/2 * * * ? 2050                                                                                                               
      extract.assets.large.job.cron: 0 0/2 * * * ? 2050                                                                                                                           
      extract.assets.batch.size.small: 5                                                                                                                           
      extract.assets.batch.size.medium: 5
      extract.assets.batch.size.large: 5    

share:
    configuration:
      share:
         existingSecrets:
         - name: alfresco-config-secret
            key: key1
            purpose: property:connector.blob.Key
```

### DIFF
--- a/templates/config-share.yaml
+++ b/templates/config-share.yaml
@@ -1,0 +1,50 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: {{ .Values.config.share.configMapName }}
+  labels:
+    {{- include "alfresco-content-services.labels" . | nindent 4 }}
+data:
+  alfresco-global.properties: |
+    {{- $alfUrl := include "alfresco-common.external.url" . }}
+    deployment.method=HELM_CHART
+    alfresco.host={{ template "alfresco-common.external.host" . }}
+    alfresco.protocol={{ template "alfresco-common.external.scheme" . }}
+    alfresco.port={{ template "alfresco-common.external.port" . }}
+    aos.baseUrlOverwrite={{ $alfUrl }}/alfresco/aos
+    csrf.filter.origin={{ $alfUrl }}
+    csrf.filter.referer={{ $alfUrl }}/.*
+    {{- if .Values.share.enabled }}
+    share.protocol={{ template "alfresco-common.external.scheme" . }}
+    share.host={{ template "alfresco-common.external.host" . }}
+    share.port={{ template "alfresco-common.external.port" . }}
+    {{- end }}
+    local.transform.service.enabled={{ index .Values "alfresco-transform-service" "enabled" }}
+    {{- with (index .Values "alfresco-transform-service") }}
+    {{- $ats_for_enterprise := and .filestore.enabled .transformrouter.enabled }}
+    transform.service.enabled={{ and .enabled $ats_for_enterprise }}
+    {{- end }}
+    {{- if index .Values "alfresco-transform-service" "enabled" }}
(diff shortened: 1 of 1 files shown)


kirananil007:
We are using custom docker images for `alfreso-repository` and `alfresco-share`. 

We got some customization to the alfresco. These are being taken care by the global properties which are getting passed into the repository and share services as far as I understand.

For alfresco-repository, the global properties are being passed into deployments via helm `values.yaml` as `additionalGlobalProperties`. 

Similarly, need a functionality for the `alfresco-share` to get the global properties being passed using helm `values.yaml`. Due to the lack of that feature currently in helm, we are baking the global-properties file for the share into the custom docker image.

alxgomz:
`alfresco-repository` have a dedicated way to pass properties so they are included in the final `alfresco-global.properties` file. This is because this file is a well-known configuration file alfresco reads on startup that's been around forever and people are used to it.
Share on the other hand will not read this file at all. Never did, and never will. Instead it uses various xml files (& a single properties file for SSO configuration). As a consequence I don't think we can merge this PR as it would be rather miss-leading for people used to share & alfresco confiuration.
Now,  if your customized Share needs to read properties fro;m a file, Share's chart - like most charts - supports support generic ways to mount configmap & secrets as files in the pod.
You can create/template your config file as a secret or a configmap and then use `extraVulmes` & `extraVolumeMounts` to get it mounted where needed.
This is more or less what we document for the metada keystore for `alfresco-repository` [here](https://alfresco.github.io/alfresco-helm-charts/charts/alfresco-repository/docs/keystores.html#the-metadata-keystore)
If your customisation of Share allows it, you can also get the properties directly from the java system properties passing them as `-Dprop=value` using `environment.CATALINA_OPTS`.
Does that help?

kirananil007:
okay got it. thanks!

PR #487, opened 2025-07-23
OPSEXP-3375: retire broken template

alxgomz:
OPSEXP-3375: Replace previous implementation with a dummy one to avoid breaking change

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,7 +86,7 @@ scheme](https://hyland.atlassian.net/wiki/spaces/TECH/pages/edit-v2/1293317631#i
   unwanted changes and push again.
 * Ensure that all the charts are getting a GA release (non-alpha) at this point.
 * Ensure the any deprecation warning is added to the global
-  [UPGRADE.md](../UPGRADE.md) file & the version about to be published indeed
-  remove the plannedremoval as stated in this same file.
+  [UPGRADES.md](UPGRADES.md) file & the version about to be published indeed
+  remove the planned removal as stated in this same file.
 * Once the PR has been merged and all charts released, you can head to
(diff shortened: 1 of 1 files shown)


UPGRADES.md
```diff
@@ -1,2 +1,6 @@
 # Deprecation
+
+- Named template `alfresco-activiti.activiti-5113.elasticsearch` is now just
+  returning the value of the `ACTIVITI_ES_REST_CLIENT_ADDRESS` entry of the
+  `environment` map. It will be retired completely in the next major release.
```
gionn, line 5:
```suggestion
  `environment` map.
```

every deprecation is going to be removed on the next major release, maybe have an introduction paragraph which explain this before the list

we should also add which chart is this otherwise is not clear

gionn (commented):
I still didn't understand why removing it, issue is really unclear

README.md
```diff
@@ -85,5 +85,8 @@ scheme](https://hyland.atlassian.net/wiki/spaces/TECH/pages/edit-v2/1293317631#i
   by updatecli according to the supported matrix, eventually revert manually
   unwanted changes and push again.
 * Ensure that all the charts are getting a GA release (non-alpha) at this point.
+* Ensure the any deprecation warning is added to the global
+  [UPGRADE.md](../UPGRADE.md) file & the version about to be published indeed
+  remove the plannedremoval as stated in this same file.
```
pmacius, line 90:
```suggestion
  [UPGRADES.md](../UPGRADES.md) file & the version about to be published indeed
  remove the planned removal as stated in this same file.
```

PR #503, opened 2025-09-15
OPSEXP-3486 Configurable postgres fsGroup

pmacius:
ref: OPSEXP-3486


charts/postgres/templates/statefulset.yaml
```diff
@@ -25,6 +25,7 @@ spec:
         {{- include "postgres.labels" $ | nindent 8 }}
     spec:
       {{- include "alfresco-common.imagePullSecrets" . | indent 6 }}
+      {{- include "alfresco-common.component-pod-security-context" . | indent 4 }}
```
gionn, line 28: you could add this also in the new elasticsearch chart so at least there is feature parity

gionn (approved):

PR #582, opened 2026-04-22
OPSEXP-3995 Bump tags for Helium release

pmacius:
OPSEXP-3995

### DIFF
--- a/charts/alfresco-repository/Chart.yaml
+++ b/charts/alfresco-repository/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: alfresco-repository
 description: Alfresco content repository Helm chart
 type: application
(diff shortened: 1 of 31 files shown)


charts/alfresco-repository/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v2
 name: alfresco-repository
 description: Alfresco content repository Helm chart
 type: application
-version: 1.3.0-alpha.0
+version: 1.3.0
```
pmacius, line 5: not sure about this

gionn: it's cool, just remember pre-release happens before a release, so to bump from pre-release to release you just remove the `-*`

gionn (approved):

PR #596, opened 2026-05-21
OPSEXP-4042: move Elasticsearch index limit settings to createIndexTemplate init container

siddavamshi4:
###OPSEXP-4042
This PR only moves two index template limit settings from configuration.search.elasticsearchProperties to initContainers.createIndexTemplate in the alfresco-repository chart.
Moved settings:

index.mapping.total_fields.limit -> initContainers.createIndexTemplate.totalFieldsLimit
index.max_result_window -> initContainers.createIndexTemplate.maxResultWindow

### DIFF
--- a/charts/alfresco-repository/templates/deployment.yaml
+++ b/charts/alfresco-repository/templates/deployment.yaml
@@ -83,22 +83,25 @@ spec:
           command: ["/bin/sh", "-c"]
           args:
             - |
+              {{- $indexName := .Values.initContainers.createIndexTemplate.indexName }}
+              {{- $maxResultWindow := .Values.initContainers.createIndexTemplate.maxResultWindow }}
+              {{- $totalFieldsLimit := .Values.initContainers.createIndexTemplate.totalFieldsLimit }}
               ES_URL="${SPRING_ELASTICSEARCH_REST_URIS%%,*}"
               until curl -sf -XPUT \
                 -H "Content-Type: application/json" \
                 -u "${SPRING_ELASTICSEARCH_REST_USERNAME}:${SPRING_ELASTICSEARCH_REST_PASSWORD}" \
                 "${ES_URL}/_index_template/{{ .Release.Name }}-{{ .Values.initContainers.createIndexTemplate.templateName }}" \
                 -d '{
-                  "index_patterns": ["{{ .Values.initContainers.createIndexTemplate.indexName }}"],
+                  "index_patterns": ["{{ $indexName }}"],
                   "template": {
                     "settings": {
                       "number_of_shards": {{ .Values.initContainers.createIndexTemplate.numberOfShards }},
                       "number_of_replicas": {{ .Values.initContainers.createIndexTemplate.numberOfReplicas }},
                       "index": {
-                        "max_result_window": {{ index .Values.configuration.search.elasticsearchProperties "index.max_result_window" | default 10000 }},
+                        "max_result_window": {{ $maxResultWindow }},
                         "mapping": {
                           "total_fields": {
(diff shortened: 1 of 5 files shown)


charts/alfresco-repository/templates/deployment.yaml
```diff
@@ -83,22 +83,25 @@ spec:
           command: ["/bin/sh", "-c"]
           args:
             - |
+              {{- $indexName := .Values.initContainers.createIndexTemplate.indexName }}
+              {{- $maxResultWindow := .Values.initContainers.createIndexTemplate.maxResultWindow }}
+              {{- $totalFieldsLimit := .Values.initContainers.createIndexTemplate.totalFieldsLimit }}
```
gionn: you can use `with` when you can restrict the values scope to a specific section, like:

```
{{- with .Values.initContainers.createIndexTemplate }}
...
        "number_of_shards": {{ .numberOfShards }},
        "number_of_replicas": {{ .numberOfReplicas }},
...
{{- end }}
```

charts/alfresco-repository/values.yaml
```diff
@@ -121,10 +121,6 @@ configuration:
       # specified as a comma-separated list (e.g.
       # `file:/path/to/file.txt,file:/path/to/file2.txt`).
       index.custom.analyzer.config.files: ""
-      # -- Maximum number of fields allowed in the search index mapping.
-      index.mapping.total_fields.limit: 7500
-      # -- Maximum number of results that can be returned by a single query.
-      index.max_result_window: 10000
```
gionn, line 127: why are we removing these?

siddavamshi4: I am just moving them to **initContainers**, should i leave them as it now or move them ?
this release only re-arranges the configuratiosn to use from initContainers instead of elasticsearchProperties
```
initContainers:
  createIndexTemplate:

Comment (siddavamshi4) [reply charts/alfresco-repository/values.yaml:127]: @alxgomz please have a look, and confirm Should i leave it as it now, or re-arrange configurations in initContainers from elasticsearchProperties

Comment (alxgomz) [reply charts/alfresco-repository/values.yaml:127]: yes keep as it is (do not remove these properties) As they are Alfresco properties which can still be used on index creation by the repo (and override the template settings)

File: charts/alfresco-repository/values.yaml [hunk 3]
```diff
@@ -376,6 +372,10 @@ initContainers:
     templateName: alfresco-template
     # -- Index name to apply the template to
     indexName: alfresco
+    # -- Maximum number of results that can be returned by a single search request.
+    maxResultWindow: 10000  # formerly index.max_result_window
+    # -- Maximum number of fields that can be created in the index mapping.
+    totalFieldsLimit: 7500  # formerly index.mapping.total_fields.limit
```
alxgomz: remove the comments and [@ignore](https://github.com/norwoodj/helm-docs/blob/a5573af096a4b526dcbc3c896c220b1714a0765b/example-charts/ignored-values-example/values.yaml#L44) the helm doc entries. The repository parameters are anyway overridding this

alxgomz (changes requested):

charts/alfresco-repository/templates/deployment.yaml
```diff
@@ -83,22 +83,26 @@ spec:
           command: ["/bin/sh", "-c"]
           args:
             - |
+              {{- with .Values.initContainers.createIndexTemplate }}
+              {{- $indexName := .indexName }}
+              {{- $maxResultWindow := .maxResultWindow }}
+              {{- $totalFieldsLimit := .totalFieldsLimit }}
```
gionn: there is no point in defining these vars, just the the `.` notation everywhere

alxgomz (approved):

PR #603, opened 2026-05-27
OPSEXP-4095 Update support.hyland.com and docs.alfresco.com references to docs.hyland.com

siddavamshi4:
OPSEXP-4095
Updated documentation references to use the consolidated Hyland documentation portal:

Replaced https://support.hyland.com/
Replaced https://docs.alfresco.com

with: https://docs.hyland.com


pmacius (approved):
Not sure about bumping the versions on doc updated but I think there is no other way. those are mentioned in values so are a part of chart anyway.

alxgomz (approved):

siddavamshi4:
> Not sure about bumping the versions on doc updated but I think there is no other way. those are mentioned in values so are a part of chart anyway.

Chart  lint has this check point to look for a version bump!

PR #637, opened 2026-07-15
OPSEXP-4173 Add alfresco-search-community chart for Elasticsearch batch indexing

gionn:
Adds a new `charts/alfresco-search-community` Helm chart deploying the Community Elasticsearch batch indexing component (`alfresco-elasticsearch-batch-indexing`), the Community counterpart to the Enterprise `alfresco-search-enterprise` connector. It is modelled on the existing `alfresco-audit-storage` single-Deployment chart and reuses `alfresco-common` helpers.

## What it deploys

The batch indexing component is a long-running singleton Spring Boot service (`replicaCount` fixed at 1) that polls the Alfresco PostgreSQL database directly and bulk-upserts nodes into Elasticsearch. Unlike the Enterprise connector it does not use ActiveMQ or the Shared File Store. Image `docker.io/alfresco/alfresco-elasticsearch-batch-indexing:5.7.0-A.5`, container port 8080.

The chart wires the database (`SPRING_DATASOURCE_*`), Elasticsearch (`SPRING_ELASTICSEARCH_REST_*`, `ELASTICSEARCH_INDEXNAME`), the ACS repository URL (`ALFRESCO_ACS_URL`), the transform config cache (`ALFRESCO_ACCEPTEDCONTENTMEDIATYPESCACHE_BASEURL`) and the content transform shared secret (`ALFRESCO_CONTENT_TRANSFORM_SHAREDSECRET`), each following the repo's value-or-`existingConfigMap`/`existingSecret` convention. A passthrough `environment` map exposes the `ALFRESCO_REINDEX_*` tunables and `JAVA_OPTS`.

## Follow-up fixes

- Added `image.internalPort` so the container's listening port no longer tracks `.Values.service.port` — previously changing the Service's exposed port also changed the container's actual port and broke Service routing.
- Fixed `configmap-es.yaml`/`secret-es.yaml` to honor `index.existingConfigMap.keys.url` and `index.existingSecret.keys.*` instead of hardcoding `SEARCH_URL`/`SEARCH_USERNAME`/`SEARCH_PASSWORD`, matching the equivalent database templates. Overriding those key names without pointing at an existing ConfigMap/Secret previously produced a Deployment referencing a key the generated resource didn't contain.
- Chart version set to `0.1.0-alpha.0` to mark it as pre-release, matching the convention used for other new/major chart versions in this repo.

## CI / testing

Registered in `test-deps.yaml` with `postgresql` + `elasticsearch` dependencies. The chart-testing probes are relaxed to a plain TCP check in `ci/default-values.yaml`, because the component needs the repository-managed Alfresco DB schema (`alf_node`, ...) to report healthy on `/actuator/health`, which is not available under chart-testing; production users keep the real health probes. Helm unit tests cover the rendered templates and README is generated via helm-docs.

Verified locally: `helm unittest`, `ct lint`, and a full `helm install --wait` on a kind cluster with the postgres and elasticsearch dependencies (pod reaches `1/1 Running`).

OPSEXP-4173

### DIFF
--- a/charts/alfresco-search-community/Chart.yaml
+++ b/charts/alfresco-search-community/Chart.yaml
@@ -1,0 +1,11 @@
+---
+apiVersion: v2
+name: alfresco-search-community
+description: A Helm chart for deploying the Alfresco Elasticsearch Community batch indexing component
+type: application
+version: 0.1.0
+appVersion: 5.7.0-A.5
+dependencies:
+  - name: alfresco-common
+    version: 5.1.0
+    repository: https://alfresco.github.io/alfresco-helm-charts/
--- a/charts/alfresco-search-community/Chart.yaml
+++ b/charts/alfresco-search-community/Chart.yaml
@@ -3,7 +3,7 @@ apiVersion: v2
 name: alfresco-search-community
 description: A Helm chart for deploying the Alfresco Elasticsearch Community batch indexing component
 type: application
-version: 0.1.0
+version: 0.1.0-alpha.0
 appVersion: 5.7.0-A.5
 dependencies:
   - name: alfresco-common
--- a/.github/workflows/lint-test.yaml
+++ b/.github/workflows/lint-test.yaml
@@ -22,7 +22,7 @@ on:
 
 env:
   TEST_ALL_CHARTS: ${{ contains(github.event.pull_request.labels.*.name, 'ci-test-all') && 'true' || 'false' }}
-  HELM_VERSION: v4.2.3
+  HELM_VERSION: v4.1.1
   CHART_TESTING_VERSION: v3.14.0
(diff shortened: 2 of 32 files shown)


charts/alfresco-search-community/Chart.yaml
```diff
@@ -0,0 +1,11 @@
+---
+apiVersion: v2
+name: alfresco-search-community
+description: A Helm chart for deploying the Alfresco Elasticsearch Community batch indexing component
+type: application
+version: 0.1.0
```
pmacius: alpha till release?

siddavamshi4 (approved):

pmacius (approved):

PR #666, opened 2026-08-12
OPSEXP-4289 mount sized /dev/shm volume to prevent shared memory exhaustion

pmacius:
## What

Add a configurable, memory-backed `/dev/shm` volume to the `postgres` chart, enabled by default (`primary.shmVolume`, `sizeLimit: 256Mi`), mounted at `/dev/shm` on the PostgreSQL container.

## Why

PostgreSQL uses POSIX dynamic shared memory (`dynamic_shared_memory_type = posix`, the default in PG 17) for parallel query execution and related operations, allocating those segments under `/dev/shm`. In a container `/dev/shm` defaults to the runtime's 64Mi tmpfs, and the chart's StatefulSet previously mounted no shm volume, so the pod ran with that 64Mi cap. Under heavier workloads — notably the ACS repository data-population job — parallel operations exhaust the 64Mi and PostgreSQL aborts with `could not resize shared memory segment ... No space left on device`, causing intermittent failures (OPSEXP-4289).

Consumers had no way to fix this from values: the chart exposed no shm/volume knob, so the only workarounds were to redirect DSM off tmpfs (`dynamic_shared_memory_type = mmap`, which trades the fast tmpfs path for disk-backed memory-mapped files) or to disable parallelism. Both are downstream band-aids. Giving PostgreSQL a right-sized `/dev/shm` fixes it at the source and keeps the faster `posix` path.

## How

The volume is a standard `emptyDir` with `medium: Memory` and a configurable `sizeLimit`, mounted at `/dev/shm`, both guarded by `primary.shmVolume.enabled`. The default 256Mi is a conservative starting point — 4x the 64Mi container default, enough to clear the observed exhaustion while sitting well below the container's existing 8Gi memory limit. A memory-backed emptyDir is charged to the pod's memory cgroup, so `sizeLimit` should stay comfortably under `primary.resources.limits.memory` and can be raised for heavier workloads. On Kubernetes with `SizeMemoryBackedVolumes` (GA), the `sizeLimit` is enforced against the tmpfs itself rather than defaulting to half of node RAM. `sizeLimit` is tunable, and the whole volume can be disabled for callers that manage shm externally.

## Testing

`helm unittest` covers the new behaviour: the volume and mount render by default, a custom `sizeLimit` propagates, and nothing renders when `shmVolume.enabled: false`. Full suite passes (8/8). `README.md` regenerated via helm-docs; chart bumped `0.7.0` → `0.8.0`.

## Rollout note

The umbrella `alfresco-content-services` chart must bump its `postgres` subchart dependency to `0.8.0` to pick this up. Downstream `mmap` workarounds can then be retired so environments regain the faster `posix` DSM path.

OPSEXP-4289

### DIFF
--- a/charts/postgres/values.yaml
+++ b/charts/postgres/values.yaml
@@ -47,6 +47,14 @@ readinessProbe:
   periodSeconds: 10
   timeoutSeconds: 3
 primary:
+  # -- Memory-backed volume mounted at /dev/shm for PostgreSQL dynamic shared
+  # memory (parallel queries). The container runtime default of 64Mi is too
+  # small for heavy workloads and causes "could not resize shared memory
+  # segment ... No space left on device". Counts against the pod memory limit,
+  # so keep sizeLimit comfortably below primary.resources.limits.memory.
+  shmVolume:
+    enabled: true
+    sizeLimit: 1Gi
   extendedConfiguration: |
     max_connections=250
     shared_buffers=512MB
--- a/charts/postgres/values.yaml
+++ b/charts/postgres/values.yaml
@@ -47,14 +47,14 @@ readinessProbe:
   periodSeconds: 10
   timeoutSeconds: 3
 primary:
-  # -- Memory-backed volume mounted at /dev/shm for PostgreSQL dynamic shared
-  # memory (parallel queries). The container runtime default of 64Mi is too
-  # small for heavy workloads and causes "could not resize shared memory
-  # segment ... No space left on device". Counts against the pod memory limit,
-  # so keep sizeLimit comfortably below primary.resources.limits.memory.
   shmVolume:
     enabled: true
-    sizeLimit: 1Gi
+    # -- Size of the memory-backed /dev/shm volume for PostgreSQL dynamic shared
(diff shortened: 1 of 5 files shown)


charts/postgres/values.yaml
```diff
@@ -47,6 +47,14 @@ readinessProbe:
...
   timeoutSeconds: 3
 primary:
+  # -- Memory-backed volume mounted at /dev/shm for PostgreSQL dynamic shared
+  # memory (parallel queries). The container runtime default of 64Mi is too
+  # small for heavy workloads and causes "could not resize shared memory
+  # segment ... No space left on device". Counts against the pod memory limit,
+  # so keep sizeLimit comfortably below primary.resources.limits.memory.
+  shmVolume:
+    enabled: true
+    sizeLimit: 1Gi
```
gionn: move comment(s) from the object to the leaf so the helm-docs are less ugly

gionn: default was 64Mi, we can start with a default which is more conservative like 256Mi?

siddavamshi4 (approved):

gionn (approved):

PR #669, opened 2026-08-24
OPSEXP-4240: fix(alfresco-search-enterprise): wait for Elasticsearch health before reindexing

siddavamshi4:
OPSEXP-4240

Ensure Search Enterprise reindexing starts only after Elasticsearch is healthy. Previously, the job waited for Repository readiness but could begin while Elasticsearch was unavailable, resulting in failed/incomplete indexing that Kubernetes incorrectly reported as successful and causing downstream Newman/DTAS search failures.
refer: https://github.com/Alfresco/acs-deployment/actions/runs/32458048808/job/96710487403#step:17:902

Add a `wait-for-elasticsearch` init container  that polls
`_cluster/health?wait_for_status=yellow` using the existing
SPRING_ELASTICSEARCH_* connection/credentials env vars, before the
reindexing container runs.

- Add `reindexing.initcontainers.waitForElasticsearch` values (image,
  resources) mirroring `waitForRepository`.
- Add unit test coverage for the new init container.
- Bump chart version to 5.3.0-alpha.0.

### DIFF
--- a/charts/alfresco-search-enterprise/templates/reindexing-job.yaml
+++ b/charts/alfresco-search-enterprise/templates/reindexing-job.yaml
@@ -120,7 +120,7 @@ spec:
           # Only start reindexing once Elasticsearch reports a healthy cluster status
           args:
             - |
-              until curl -sf -u "${SPRING_ELASTICSEARCH_REST_USERNAME}:${SPRING_ELASTICSEARCH_REST_PASSWORD}" \
+              until curl -sf --connect-timeout 5 --max-time 10 -u "${SPRING_ELASTICSEARCH_REST_USERNAME}:${SPRING_ELASTICSEARCH_REST_PASSWORD}" \
                 "${SPRING_ELASTICSEARCH_REST_URIS}/_cluster/health?wait_for_status=yellow&timeout=5s" -o /dev/null
               do
                 echo "Waiting for Elasticsearch to become healthy at ${SPRING_ELASTICSEARCH_REST_URIS} ..."
--- a/charts/alfresco-search-enterprise/templates/reindexing-job.yaml
+++ b/charts/alfresco-search-enterprise/templates/reindexing-job.yaml
@@ -125,13 +125,10 @@ spec:
             {{- include "alfresco-search-enterprise.config.spring.envCredentials" $ | nindent 12 }}
             {{- end }}
           command: ["/bin/sh", "-c"]
-          # Only start reindexing once Elasticsearch reports a healthy cluster status.
-          # Only the first URI is checked when multiple are configured, matching
-          # alfresco-repository's initContainers.createIndexTemplate pattern.
           args:
             - |
               ES_URL="${SPRING_ELASTICSEARCH_REST_URIS%%,*}"
-              check_cluster_health() {
+              until
               {{- if $iam }}
                 awscurl \
                   --fail-with-body \
@@ -145,8 +142,6 @@ spec:
                   -u "${SPRING_ELASTICSEARCH_REST_USERNAME}:${SPRING_ELASTICSEARCH_REST_PASSWORD}" \
               {{- end }}
                   "${ES_URL}/_cluster/health?wait_for_status=yellow&timeout=5s"
-              }
-              until check_cluster_health
               do
                 echo "Waiting for Elasticsearch to become healthy at ${ES_URL} ..."
                 sleep 5
--- a/charts/alfresco-search-enterprise/templates/reindexing-job.yaml
+++ b/charts/alfresco-search-enterprise/templates/reindexing-job.yaml
@@ -106,27 +106,54 @@ spec:
               done
               echo 'Alfresco is ready, delay reindexing to give a chance to fully initialise.'
               sleep 30
+        {{- if .Values.reindexing.initcontainers.waitForElasticsearch.enabled }}
+        {{- with .Values.reindexing.initcontainers.waitForElasticsearch }}
+        {{- $mode := (.auth.mode | default "basic") }}
+        {{- if not (has $mode (list "basic" "iam")) }}
+        {{- fail (printf "reindexing.initcontainers.waitForElasticsearch.auth.mode must be \"basic\" or \"iam\", got %q" $mode) }}
+        {{- end }}
+        {{- $iam := eq $mode "iam" }}
         - name: wait-for-elasticsearch
-          {{- with .Values.reindexing.initcontainers.waitForElasticsearch }}
-          image: {{ printf "%s:%s" .image.repository .image.tag | quote }}
-          imagePullPolicy: {{ .image.pullPolicy }}
-          resources: {{- toYaml .resources | nindent 12 }}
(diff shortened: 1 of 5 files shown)


charts/alfresco-search-enterprise/templates/reindexing-job.yaml
```diff
@@ -106,7 +106,27 @@ spec:
...
+          {{- include "alfresco-common.component-security-context" .Values | indent 8 }}
+          env:
+            {{- include "alfresco-search-enterprise.config.spring.es.env" $ | nindent 12 }}
+            {{- include "alfresco-search-enterprise.config.spring.envCredentials" . | nindent 12 }}
+          command: ["/bin/sh", "-c"]
+          # Only start reindexing once Elasticsearch reports a healthy cluster status
+          args:
+            - |
+              until curl -sf --connect-timeout 5 --max-time 10 -u "${SPRING_ELASTICSEARCH_REST_USERNAME}:${SPRING_ELASTICSEARCH_REST_PASSWORD}" \
+                "${SPRING_ELASTICSEARCH_REST_URIS}/_cluster/health?wait_for_status=yellow&timeout=5s" -o /dev/null
```
gionn: SPRING_ELASTICSEARCH_REST_URIS naming looks containing something which make this work only when there is only one uri

charts/alfresco-search-enterprise/templates/reindexing-job.yaml
```diff
@@ -106,7 +106,27 @@ spec:
...
+          {{- end }}
+          {{- include "alfresco-common.component-security-context" .Values | indent 8 }}
+          env:
+            {{- include "alfresco-search-enterprise.config.spring.es.env" $ | nindent 12 }}
+            {{- include "alfresco-search-enterprise.config.spring.envCredentials" . | nindent 12 }}
+          command: ["/bin/sh", "-c"]
+          # Only start reindexing once Elasticsearch reports a healthy cluster status
+          args:
+            - |
+              until curl -sf --connect-timeout 5 --max-time 10 -u "${SPRING_ELASTICSEARCH_REST_USERNAME}:${SPRING_ELASTICSEARCH_REST_PASSWORD}" \
```
gionn: plain curl won't work with iam auth, see https://github.com/Alfresco/alfresco-helm-charts/blob/main/charts/alfresco-repository/templates/deployment.yaml#L92

charts/alfresco-search-enterprise/values.yaml
```diff
@@ -291,6 +291,15 @@ reindexing:
         limits:
           cpu: "250m"
           memory: "20Mi"
+    waitForElasticsearch:
```
gionn, line 294: should have an enabled and default to false because it's risky to introduce

gionn (changes requested):
pr description seems partial and doesn't support why we need this

charts/alfresco-search-enterprise/templates/reindexing-job.yaml
siddavamshi4: Thanks for the reference

gionn (changes requested):
if we are going to validate it's a reindex app issue (not properly failing when the reindex fails) then we are not gonna need this

charts/alfresco-search-enterprise/values.yaml
```diff
@@ -291,6 +291,36 @@ reindexing:
...
+      # -- Wait for Elasticsearch cluster health before starting reindexing. Disabled by default. Only the first URI in `SPRING_ELASTICSEARCH_REST_URIS` is checked.
+      enabled: false
+      auth:
+        # -- Authentication mode for the health-check request: `basic` (HTTP basic auth) or `iam` (AWS SigV4, signed with the pod's AWS identity via IRSA)
+        mode: basic
+        aws:
+          # -- AWS region of the OpenSearch domain; required when mode is `iam`
+          region: null
+          # -- AWS service name for SigV4 signing (defaults to `es` for managed OpenSearch/Elasticsearch)
+          service: es
```
alxgomz, line 304: Check At least https://github.com/Alfresco/acs-deployment/blob/master/docs/helm/examples/with-opensearch-iam-auth.md#helm-values where we already document how to pass the service.
No need for a new value here, that just makes the configuration process more inconsistent.

We could even hard code it IMHO

charts/alfresco-search-enterprise/values.yaml
```diff
@@ -291,6 +291,36 @@ reindexing:
...
           memory: "20Mi"
+    waitForElasticsearch:
+      # -- Wait for Elasticsearch cluster health before starting reindexing. Disabled by default. Only the first URI in `SPRING_ELASTICSEARCH_REST_URIS` is checked.
+      enabled: false
+      auth:
+        # -- Authentication mode for the health-check request: `basic` (HTTP basic auth) or `iam` (AWS SigV4, signed with the pod's AWS identity via IRSA)
+        mode: basic
+        aws:
+          # -- AWS region of the OpenSearch domain; required when mode is `iam`
+          region: null
```
alxgomz, line 302: Check At least https://github.com/Alfresco/acs-deployment/blob/master/docs/helm/examples/with-opensearch-iam-auth.md#helm-values where we already document how to pass the ES region. 
No need for a new value here, that just makes the configuration process more inconsistent.

charts/alfresco-search-enterprise/values.yaml
```diff
@@ -291,6 +291,36 @@ reindexing:
...
+          # -- AWS service name for SigV4 signing (defaults to `es` for managed OpenSearch/Elasticsearch)
+          service: es
+      # -- Init container image per auth mode; the entry matching `auth.mode` is used
+      images:
```
alxgomz, line 306: You can reduce complexity by using `curl` in both cases as it supports AWS sign v4 for quite some time already.
You will need to use shell env to have proper env var expansion available though.

alxgomz (changes requested):

charts/alfresco-search-enterprise/templates/reindexing-job.yaml
```diff
@@ -106,7 +106,49 @@ spec:
...
+                  -o /dev/null \
+                  --service {{ .auth.aws.service | quote }} \
+                  --region {{ required "reindexing.initcontainers.waitForElasticsearch.auth.aws.region is required when auth.mode is iam" .auth.aws.region | quote }} \
+              {{- else }}
+                curl \
+                  -sf --connect-timeout 5 --max-time 10 \
+                  -o /dev/null \
+                  -u "${SPRING_ELASTICSEARCH_REST_USERNAME}:${SPRING_ELASTICSEARCH_REST_PASSWORD}" \
+              {{- end }}
+                  "${ES_URL}/_cluster/health?wait_for_status=yellow&timeout=5s"
```
alxgomz, line 144: In the light of our latest investigations I would even test for the `indexName` existance, to avoid ES to create index with dynamic fields on first requests

alxgomz (changes requested):

